PR #14, opened 2017-07-21
No strangers

maxf:
Because we want to mandate HTTP auth, we set defaults to be long numbers. There's probably a clearer way to do it. Suggestions welcome.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,3 +1,6 @@
 # Find Data Beta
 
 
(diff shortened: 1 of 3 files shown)


govuklaurence:
Might be best to ditch the random numbers and just fail if the env vars aren't set

maxf:
@govuklaurence : fail how?

maxf:
How's that @govuklaurence ?

PR #32, opened 2017-07-25
Rough cut dataset convencience object

govuklaurence:
OK! All methods take same params as in `Elasticsearch::Client`.

This satisfies initial reqs for #30.

## Get a dataset:
```
Dataset.get(id: 124)
```

## Search for datasets
```
ds = Dataset.search(q: 'test')
ds.datasets # dataset objects
ds.raw # raw results
```

## Custom query (returns raw results for any method on client)
```
ds = Dataset._query(:search, q: 'test')
```

### DIFF
--- a/app/models/dataset.rb
+++ b/app/models/dataset.rb
@@ -1,0 +1,51 @@
+class DatasetSearchResponse
+  include ActiveModel::Model
+  attr_accessor :datasets, :raw
+  # This class can be improved by porting over more metadata from the raw response
+end
+
+class Dataset
+  include ActiveModel::Model
+  attr_accessor :name, :title, :summary, :description,
+       :location1, :location2, :location3,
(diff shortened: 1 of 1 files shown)


app/models/dataset.rb
```diff
@@ -0,0 +1,51 @@
...
+  include ActiveModel::Model
+  attr_accessor :name, :title, :summary, :description,
+       :location1, :location2, :location3,
+       :licence, :licence_other, :frequency,
+       :published_date, :updated_at, :created_at,
+       :harvested, :uuid, :organisation, :datafiles,
+       :inspire_dataset, :json,
+       :_index, :_type, :_id, :_score, :_source,
+       :_version
+  # TODO: subclass for datafiles
```
rossjones: Also organisation (will be a publisher page at some point).

PR #42, opened 2017-07-26
Dataset view

hannako:
- creates all partials for the dataset show page
- temporary objects passed in from dataset controller for time being


app/controllers/application_controller.rb
```diff
@@ -1,9 +1,8 @@
 class ApplicationController < ActionController::Base
-  before_action :authenticate
+  # before_action :authenticate
```
rossjones: Is there a reason this is disabled?  We need this (for now) on heroku so that it doesn't accidentally get spidered by google.

app/views/datasets/_additional_info.html.erb
```diff
@@ -0,0 +1,13 @@
+<section class="info">
```
rossjones: This should come from the `notes` field of the elastic search result.

app/views/datasets/_feedback.html.erb
```diff
@@ -0,0 +1,62 @@
+<section id="feedback">
```
rossjones: Do we have a ticket to send this info somewhere?

PR #48, opened 2017-07-28
Search page views

### LINKED ISSUE #35: Search page views
#5

govuklaurence:
Closes #35


app/controllers/search_controller.rb
```diff
@@ -0,0 +1,21 @@
+class SearchController < ApplicationController
+  def search
+    @query = params["q"]
+    datasets = Dataset.search({ q: @query })
```
hannako: maybe this guy should be called `results`

app/controllers/search_controller.rb
```diff
@@ -0,0 +1,21 @@
...
+    @results = datasets.datasets
```
hannako: and this guy would be `@datasets = results.datasets`

app/controllers/search_controller.rb
govuklaurence: Agreed, but i'll catch them in a seperate pass

PR #55, opened 2017-08-01
Elastic Config & Cleanup

TobyRet:


spec/spec_helper.rb
```diff
@@ -2,9 +2,18 @@
 SimpleCov.start
 
 RSpec.configure do |config|
+  config.before(:each) do
+    ELASTIC.indices.create index: "datasets-test"
```
rossjones: Might that fail if I abort the test?

govuklaurence: Agreed, should try to delete before as well.

TobyRet: Thanks @govuklaurence @rossjones  - will push changes shortly

PR #62, opened 2017-08-02
Display time series data

TobyRet:


app/helpers/datasets_helper.rb
```diff
@@ -1,82 +1,105 @@
...
+  def dataset_location(dataset)
+    locations(dataset).empty? ? NO_MORE['default'] : locations(dataset)
+  end
+
+  def expected_location_class_for(dataset)
+    NO_MORE['default'] if locations(dataset).empty?
+  end
+
+  def timeseries_data?(dataset)
+    dataset['datafiles'].map {|file| Time.parse(file['start_date']).year}.uniq
```
hannako: I hadn't finished this guy, it is just returning an array of years at the mo. If it's empty or has only one date in it then it's not time series?

spec/features/dataset_show_page_spec.rb
```diff
@@ -1,125 +1,107 @@
...
-     }]
-
-     it "uses the updated_at field to calculate next update" do
-       dataset = create_dataset("Lovely data", "annual", datafiles_no_enddate)
-       index_and_visit(dataset)
-       expect(page).to have_content("Expected update: 31 August 2017")
-     end
-  end
+  describe 'Datafiles are not present' do
+    context 'Sections' do
```
hannako: Convention is context is for state and describe is for things. I think they're the same method - but just fyi.

spec/helpers/datasets_helper_spec.rb
```diff
@@ -1,57 +1,88 @@
...
+                  'name' => 'Dataset 3',
+                  'url' => 'https://good_data.co.uk',
+                  'start_date' => '2015-09-24',
+                  'end_date' => nil,
+                  'updated_at' => '2015-08-31T14:40:57.528Z',
+                  'start_year' => '2015'
+                 }]
+  }
+
+  it 'Should group time series data by year' do
```
hannako: I think the [convention ](http://www.betterspecs.org/#expect)is that you don't use should.

PR #79, opened 2017-08-04
Add related datasets to datasets view page

TobyRet:
Appreciate any comments on what would be appropriate testing here.
Would an E2E test suffice?


TobyRet:
[Related datasets](https://trello.com/c/gDvb6iou/310-related-datasets)

govuklaurence:
E2E test would be most useful, yes. Just check that related links is populated on a given dataset

Issue #84, opened 2017-08-07
Fix filtering code


govuklaurence:
Errors occur on clicking a publisher name in search results, and hitting dataset title on results page

Issue #52, opened 2017-07-31
Show all datasets when no search term provided.


rossjones:

govuklaurence:
Fixed by #84

PR #88, opened 2017-08-07
Fix null search error

govuklaurence:


govuklaurence:
Fixes #84 and #52

hannako:
- This fixes the 500 when you click on the publisher name

- It doesn't fix the 500 when you click on the dataset title

- Shall I approve and merge in and you can create new ticket for the latter?

govuklaurence:
Yeah, please approve and merge and split out a seperate ticket, as the other error is unrelated to these changes.

PR #95, opened 2017-08-08
Display datasets tests

TobyRet:
Note that dynamic error pages have now been created (404 and 500), Rails defaults have been removed. Have added dummy copy for now.


app/controllers/datasets_controller.rb
```diff
@@ -4,9 +4,16 @@ class DatasetsController < ApplicationController
   include DatasetsHelper
 
   def show
+    begin
+      @dataset = current_dataset
+      raise 'Metadata missing' if @dataset['title'].empty?
+
+    rescue
+      render :template => "errors/not_found", :status => 404
+    end
```
maxf: It would be nice to have a different 404 body message, to differentiate real 404s from metadata-missing 404s. Or we can find another http response code that matches the semantics.

app/views/datasets/_meta_data.html.erb
```diff
@@ -13,15 +12,15 @@
             <dt>Last updated:</dt>
             <dd><%= format(@dataset["updated_at"]) %></dd>
             <dt>Expected update:</dt>
-              <% if @dataset["datafiles"].any? %>
-                <dd class="<%= expected_update_class_for(@dataset['frequency'])%>">
+            <% if @dataset["datafiles"].any? %>
+                <dd class="<%= expected_update_class_for(@dataset['frequency']) %>">
```
maxf: indentation seems wrong

maxf: and in other places below too

app/views/datasets/show.html.erb
```diff
@@ -19,7 +19,7 @@
       <%= render "supporting_docs" %>
   <% end %>
 
-  <% if @dataset['organisation']['contact_name'] || @dataset['organisation']['contact_email'] %>
+  <% if @dataset['organisation']['contact_email'].present? %>
```
hannako: This seems sensible, but maybe need to run passed gurinder/blanca? I think there is a general vibe that is if there's info display it - perhaps it's useful to know the name of the contact person?

app/views/errors/internal_server_error.html.erb
```diff
@@ -0,0 +1,5 @@
+<main id="content" role="main">
```
maxf: Should you set the page title here? Usually something like "Page not found - 404 - data.gov.uk"

app/views/errors/internal_server_error.html.erb
```diff
@@ -0,0 +1,5 @@
...
+  <div class="grid-row">
+    <h1 class="heading-large">Something broke!</h1>
+  </div>
+</main>
```
maxf: We normally add line-breaks at the end of files

config/routes.rb
```diff
@@ -1,4 +1,7 @@
 Rails.application.routes.draw do
+  match "/404", :to => "errors#not_found", :via => :all
+  match "/500", :to => "errors#internal_server_error", :via => :all
+
```
maxf: The problem with 500 pages defined using controllers is that if the whole app breaks it can't even be rendered. That's why there's 404.html and 500.html in /public by default. We're likely to have an nginx server in front of the app anyway, which will serve its own static 500 page if the application is down.

TobyRet: Yep that right - the new error pages are created dynamically and we will need to have something on nginx incase the app goes down

app/views/errors/internal_server_error.html.erb
TobyRet: Interesting - let's discuss

app/views/errors/internal_server_error.html.erb
TobyRet: Currently the page title is 'Something broke'.
Waiting content from Gurinda

app/views/datasets/_meta_data.html.erb
TobyRet: mmm ... this appears to be default settings via the Intellij ruby plugin
Its two spaces but this increases to four when you start a new line after executing ruby code <% ... %>. This can be changed though I'm sure. 
Let's discuss

app/controllers/datasets_controller.rb
TobyRet: I'm guessing from the user perspective that the nature of the error will not be of interest but agree that logging the type of error is required. Currently, the conditions under which information is displayed or not are still to be finalised ... This is a first pass on the back of an initial discussion with Jonathon. 

Are there any plans to set up a logging service like https://logentries.com/ so that we can can keep track of errors?

app/views/datasets/show.html.erb
TobyRet: Yes agree. This is just a first pass on the back of initial discussion with Jonathon. Acceptance criteria here - https://trello.com/c/aOhd5nFJ/332-test-cases-dataset-view-page

TobyRet: Yep need to check with them. I will raise a ticket

TobyRet:
[Test cases: Dataset view page](https://trello.com/c/aOhd5nFJ/332-test-cases-dataset-view-page)

app/controllers/datasets_controller.rb
maxf: We're going to use log.it, but I haven't looked into it yet.

app/views/datasets/_meta_data.html.erb
TobyRet: I see it's a setting called 'continuation indent' which is currently set to 4 spaces. If I set to 2 will this be consistent with your IDE?

app/views/errors/internal_server_error.html.erb
TobyRet: @maxf - have fixed settings in my IDE

PR #97, opened 2017-08-10
Get all publishers

hannako:
- pulls in list of publishers from Publish at the moment. 
- removes redundant prototype kit code from govuk.js and replaces const with var
- works fine, but we can refactor to use an Elastic search index at a later date. I'll make a new ticket.


hannako:
[MVP feature to display all datasets published by organisation](https://trello.com/c/qdXRxI0M/31-mvp-feature-to-display-all-datasets-published-by-organisation)

PR #99, opened 2017-08-10
Link to datasets using friendly urls

TobyRet:
**NOTE**
A new Elasticsearch mapping has been added (please see - https://github.com/datagovuk/publish_data_beta/pull/321). Local and production elasticsearch indexes will need to be deleted and reindexed

I have started using the Elasticsearch DSL for querying - I was running into some problems with Elasticsearch Model and believe this is because the latter only handles simple queries.

Hope that is ok

### DIFF
--- a/app/controllers/datasets_controller.rb
+++ b/app/controllers/datasets_controller.rb
@@ -6,34 +6,26 @@ class DatasetsController < ApplicationController
   def show
     begin
       @dataset = current_dataset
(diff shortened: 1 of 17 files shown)


hannako:
[Use name when linking to dataset pages.](https://trello.com/c/aZ42z2uB/344-use-name-when-linking-to-dataset-pages)

PR #113, opened 2017-08-14
Find Beta - search page parity with Alpha

TobyRet:
Fixes for search page filtering
- filter by location
- filter by publisher
- apply one or more filters as required


TobyRet:
[Find Beta - search page parity with Alpha](https://trello.com/c/RqK4pdit/268-find-beta-search-page-parity-with-alpha)

PR #114, opened 2017-08-14
Results page layout

maxf:


maxf:
@TobyRet I fixed the first issue. The second is I think outside of the scope of this ticket so will create another one.

PR #125, opened 2017-08-15
Maxf/no results

maxf:


hannako:
[Dealing with zero search results](https://trello.com/c/WmJIoTis/301-dealing-with-zero-search-results)

PR #135, opened 2017-08-16
Render Previews on Find

govuklaurence:
* Adds a route to find that fetches and shows previews
* Hooks up preview links next to files
* Basic error handling
* New route
* Initialises a constant PUBLISH_URL that reads from an ENV VAR of the same name
* Two basic tests

Now with cleaner history

### DIFF
--- a/config/initializers/preview_fetch_initializer.rb
+++ b/config/initializers/preview_fetch_initializer.rb
@@ -1,7 +1,7 @@
 require 'rest-client'
 require 'uri'
 
-PUBLISH_URL = ENV['PUBLISH_URL']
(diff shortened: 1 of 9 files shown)


config/initializers/preview_fetch_initializer.rb
```diff
@@ -1,7 +1,7 @@
 require 'rest-client'
 require 'uri'
 
-PUBLISH_URL = ENV['PUBLISH_URL']
+PUBLISH_URL = ENV['PUBLISH_URL'] || "https://publish-data-beta.herokuapp.com"
```
TobyRet: @govuklaurence - should we move this to yml config now or later?

govuklaurence: It's ok to stay here for now, but we can add a card to yamlize it

PR #141, opened 2017-08-18
Fix elasticsearch yaml configuration and importing of env vars

TobyRet:


govuklaurence:
Please deploy to heroku to verify this is working first, then will merge

PR #171, opened 2017-08-21
Boost location and organisation in related dataset results

TobyRet:
Keen for comments on this. 

This is a fine tuning of the related datasets results - those datasets that have the same publisher name or location will be scored higher. 

Note that this is not a filtering query and the results are determined based on scoring (hence there is a chance you could have a related dataset returned from another publisher, however this should be scored lower and appear lower down in the results) - hence the request for feedback.

If we want to explicitly filter datasets by any query filters then we would need to pass the filter values to the datasets controller - I'm not sure whether this is more valuable based on the limited feedback obtained so far

Code climate hates me @govuklaurence

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,3 +3,4 @@
 ## 2017-08-11
 * Added changelog
 * Added contributors list and contribution instructions
+* Tweak location and organisation scoring in search results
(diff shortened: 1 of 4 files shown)


TobyRet:
[Related datasets have do not take query filters into account](https://trello.com/c/m9Ga3rvM/374-related-datasets-have-do-not-take-query-filters-into-account)

PR #174, opened 2017-08-21
Use twitter-typeahead instead of accessible-autocomplete

maxf:
Rationale: in order to make accessible typeahead work without JS, it has to be a dropdown.
We have hundreds of organisations and locations, so a dropdown would be unusable. Instead,
a-t lets us use a normal text input and progressively enhance it with typeahead


app/assets/javascripts/govuk.js
```diff
@@ -24,27 +24,10 @@ $(document).ready(function () {
...
-    accessibleAutocomplete({
-      element: locationInput,
-      id: 'location',
-      name: locationInput.getAttribute('name'),
-      source: suggest_locations,
-      showAllValues: true, dropdownArrow: () => '',
-    })
-  }
 
+  addAutocomplete('#publisher', 'publisher', publisherTitles)
```
hannako: Do you think we should split this out as a module and put it in its own file in `app/assets/javascripts/`

app/assets/stylesheets/components.scss
```diff
@@ -41,6 +41,45 @@
...
+    background: $page-colour;
+    overflow-x: hidden;
+    overflow-y: auto;
+    max-height: 150px;
+  }
+  .tt-dropdown-menu {
+    width: 100%;
+  }
+  .tt-highlight {
+    text-decoration: underline;
```
hannako: I think we are missing ` cursor: pointer;` here

hannako:
Also need to add a line to the changelog

maxf:
@hannako : all done, I think.

PR #179, opened 2017-08-23
Sort search results

TobyRet:
This PR wires up the sort drop down menu to the back end.
Also provides working functionality for users with no javascript.


TobyRet:
[“Apply filters” applies sorting as well, it should only apply to filters](https://trello.com/c/snTnmZtb/371-apply-filters-applies-sorting-as-well-it-should-only-apply-to-filters)

app/assets/javascripts/find-data.js
```diff
@@ -0,0 +1,13 @@
+function sortDatasets() {
```
maxf: This file's name is confusing. And there are no comments to explain what it does.

app/assets/javascripts/find-data.js
```diff
@@ -0,0 +1,13 @@
...
+    document.location.href = newLocation()
+
+    function newLocation() {
```
maxf: I don't think that creating a newLocation function is necessary. It brings the risk of things breaking is the declaration ever changes to ` var newLocation = function...` and stop the hoisting of the function. Better to just write `document.location.href = function () {...`

app/assets/javascripts/find-data.js
```diff
@@ -0,0 +1,13 @@
...
+        var currentLocation = document.location.href
+        var sortParam = "&sort="
```
maxf: We use 2 spaces for indentation.

app/assets/stylesheets/govuk-overrides.scss
```diff
@@ -49,3 +49,22 @@
     }
   }
 }
+
+// ______________________________________________________
+
+// Search
+
+.apply-sort {
+  float: right;
```
maxf: This doesn't belong in this file, which is originally intended for modifications to the standard GOV.UK style. This rather belongs to components.scss

app/views/search/_sort.html.erb
```diff
@@ -1,9 +1,14 @@
+<div class='apply-sort hidden'>
+  <input type="submit" value="Apply sorting"/>
```
maxf: Did you mean `js-hidden` ?

app/views/search/_sort.html.erb
```diff
@@ -1,9 +1,14 @@
...
+</div>
+
 <div class="dgu-dropdown">
   <label>
     <span class="visuallyhidden">Sorting type</span>
-    <select name="sort">
+    <select id='sort-datasets' name="sort" onchange="sortDatasets()">
       <option <%= 'selected' if @sort!='recent' %> value="best">Best match</option>
```
maxf: Why doesn't `onchange="this.form.submit()"` work?

app/assets/javascripts/find-data.js
maxf: I recommend (and probably eventually mandate) to run all js files against jslint before committing.

Issue #145, opened 2017-08-18
Most recent filter does not work, last updated is set incorrectly


govuklaurence:

govuklaurence:
![](https://github.trello.services/images/mini-trello-icon.png) [Gauntlet Bug Fixes](https://trello.com/c/i3BEJXyD/353-gauntlet-bug-fixes)

nicholsj:
![](https://github.trello.services/images/mini-trello-icon.png) [Most recent filter not working - all datasets have last updated set to the same date - Aug 2017](https://trello.com/c/9u6o89zd/375-most-recent-filter-not-working-all-datasets-have-last-updated-set-to-the-same-date-aug-2017)

PR #180, opened 2017-08-23
Use new dataset date attribute (last_updated_at instead of updated_at)

hannako:
This is the front end part of the fix for #145 . We are reading from the new field `last_updated_at` which contains the value stored in `metadata_modified` on the legacy dataset

### DIFF
--- a/app/helpers/query_builder.rb
+++ b/app/helpers/query_builder.rb
@@ -65,7 +65,7 @@ module QueryBuilder
 
     case sort_param
       when "recent"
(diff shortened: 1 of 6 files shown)


hannako:
[Most recent filter not working - all datasets have last updated set to the same date - Aug 2017](https://trello.com/c/9u6o89zd/375-most-recent-filter-not-working-all-datasets-have-last-updated-set-to-the-same-date-aug-2017)

govuklaurence:
Rebased to master

PR #181, opened 2017-08-23
Download label if html

hannako:
@govuklaurence 
If we run the [linkchecker](https://github.com/datagovuk/publish_data_beta/blob/393c4664a3c88672d0d4f8f7ac2d830dfba91502/lib/tasks/check.rake#L22) rake task on publish before we reindex, then I think all of the empty `datafile.format`s will be [filled in nicely](https://github.com/datagovuk/publish_data_beta/blob/393c4664a3c88672d0d4f8f7ac2d830dfba91502/lib/util/linkchecker.rb#L40).


hannako:
[“Download” label for html-type data files doesn’t make sense. Change to “View” if link is to html](https://trello.com/c/Kzj3bLIl/364-download-label-for-html-type-data-files-doesnt-make-sense-change-to-view-if-link-is-to-html)

PR #185, opened 2017-08-24
Fix HTML escaping

govuklaurence:


app/views/datasets/_meta_data.html.erb
```diff
@@ -31,7 +31,7 @@
         <h3 class="heading-small">
           Summary
         </h3>
-        <p class="summary" style="max-height: 250px; overflow: hidden"> <%= strip_tags(@dataset.summary) %> </p>
+        <p class="summary" style="max-height: 250px; overflow: hidden"> <%= unescape(@dataset.summary) %> </p>
       </div>
```
maxf: How about other fields? Potentially all text fields could have escaped characters, as people have pasted them

PR #187, opened 2017-08-24
Add facility for generating passwords for private beta users

govuklaurence:
This adds a 'password' generator for private beta users. It's not actually a password at all, it's just a hash that's easy to read.

You can get the 'password' for any username by running:
`rails user:generate[username]`

This will be taken out before public beta launches - it's not meant to secure the beta, just keep google and random others from accidentally getting to it. This is not a secure facility and should not be considered as such.

Example usage (run with no salt):
```
$ rails user:generate[laurence]
uniform india juliet eight echo papa delta zulu
```

You should set a salt when running in a staging or prod env (so that people can't use their own machines to generate passwords):
```
$ export PRIVATE_BETA_USER_SALT=my_super_secret_salt
```

You can blacklist users as well
```
$ export PRIVATE_BETA_USER_BLACKLIST=laurence;max;jess
```
and those users won't be able to log in, even with the right 'password'.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,3 +22,4 @@
 
 * Create empty state for dataset location and last updated (#151)
 * Download or View depending if link is HTML or other (#147)
+* Add a login facility for private beta users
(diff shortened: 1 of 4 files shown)


govuklaurence:
[Build landing page with consent information for beta.gov.uk](https://trello.com/c/d22PFOI0/395-build-landing-page-with-consent-information-for-betagovuk)

PR #201, opened 2017-08-30
Add required preview links and templates

govuklaurence:
This is a bit messy at the moment, I plan to refactor and add some tests tomorrow. It also needs some more support from publish API for name, number of rows, etc.

### DIFF
--- a/app/controllers/datasets_controller.rb
+++ b/app/controllers/datasets_controller.rb
@@ -20,10 +20,10 @@ class DatasetsController < ApplicationController
   end
 
   def preview
-    @preview = RestClient.get(preview_url(params[:file_id]))
-    render json: @preview
+    @preview = JSON.parse(RestClient.get(preview_url(params[:file_id])))
+    @content_type = @preview['content']['type']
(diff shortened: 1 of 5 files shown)


hannako:
Awaiting changes on publish

app/controllers/datasets_controller.rb
```diff
@@ -20,10 +20,11 @@ def show
...
-    @preview = RestClient.get(preview_url(params[:file_id]))
-    render json: @preview
-  rescue
-    render json: { "error": "No preview available" }, status: 404
+    @preview = JSON.parse(RestClient.get(preview_url(params[:file_id])))
+    @content_type = @preview.fetch('content', {}).fetch('type', '')
+    @content_type = @content_type.upcase
+  rescue => e
+    logger.warn("Error while displaying preview: #{e}")
   end
```
maxf: Does that rescue result in a 500 page?

govuklaurence: No - it results in a 'preview can't be displayed' page.

PR #211, opened 2017-08-31
Fix location filter

TobyRet:
Use a 'match' query instead of 'filter' query. The latter requires fields to not be analyzed in order to work


hannako:
@TobyRet I spoke to Laurence, and his view that we go with this behaviour for now. So I'm merging :)

PR #218, opened 2017-09-01
Copy changes and fix for broken tests

TobyRet:


TobyRet:
[Copy changes](https://trello.com/c/uw2kHzQq/414-copy-changes)

app/views/search/search.html.erb
```diff
@@ -1,15 +1,7 @@
-<% content_for :page_title do %>
-  Results for "<%= @query %>" - Find Data Alpha
-<% end %>
-
```
maxf: Why has the page title disappeared?

TobyRet: @maxf - I believe this is dead code copies over by accident from alpha
It's removal does not have a side effect

TobyRet: <% content_for :page_title do %>
  Data.gov.uk - Find Data
<% end %>

is set in application.html.erb

TobyRet: Ah I see now, my bad. Will change

maxf: That code sets the `<title>` for the search results page (ie, 'search results for ...'. If we don’t include it, it’ll have the default title for all pages, which is an accessibility issue.

PR #219, opened 2017-09-01
Preview styling

hannako:
@govuklaurence 

This PR

- Styles the preview page NB line count is still to be added. See [here](https://github.com/datagovuk/find_data_beta/pull/224/files)
- Restructures table as per https://trello.com/c/u90lefkO/423-datafiles-table-styling

new layout:

screen shot 2017-09-04 at 22 37 20

- Creates a back to dataset link based on the id preview['meta']['dataset_id']
- NB as discussed on slack, there is a problem with the IDs that are coming over from publish:
  - The preview link takes you to a preview that doesn't belong to that datafile.
  - The back to dataset link takes you to a dataset that is not the parent of that datafile.

- This is safe to be merged into master now, no previews will be shown as `not available` will show for all non-html file types

### DIFF
--- a/app/helpers/datasets_helper.rb
+++ b/app/helpers/datasets_helper.rb
@@ -1,5 +1,4 @@
 module DatasetsHelper
-# include Rails.application.routes.url_helpers
 
   FREQUENCIES = {
       'annual' => {years: 1},
@@ -73,12 +72,12 @@ module DatasetsHelper
     (datafile["format"].blank? || datafile["format"] == 'HTML') ? 'View' : 'Download'
   end
 
-  def preview_exists?(datafile)
+  def preview_exists?(url)
     begin
-    response = RestClient.get(preview_url(datafile['id']))
-    !response.body.empty?
+      response = RestClient.get(url)
+      !response.body.empty?
     rescue
-      false
+      'no preview available'
     end
   end
   private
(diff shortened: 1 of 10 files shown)


app/helpers/datasets_helper.rb
```diff
@@ -1,4 +1,5 @@
 module DatasetsHelper
+# include Rails.application.routes.url_helpers
```
govuklaurence: Rails should include this by default -- remove

app/helpers/datasets_helper.rb
```diff
@@ -72,6 +73,14 @@ def format_button(datafile)
     (datafile["format"].blank? || datafile["format"] == 'HTML') ? 'View' : 'Download'
   end
 
+  def preview_exists?(datafile)
+    begin
+    response = RestClient.get(preview_url(datafile['id']))
```
govuklaurence: Indentation required

app/helpers/datasets_helper.rb
```diff
@@ -72,6 +73,14 @@ def format_button(datafile)
...
+    !response.body.empty?
+    rescue
```
govuklaurence: Try and use a more specific rescue if possible

hannako:
[Datafiles table styling](https://trello.com/c/u90lefkO/423-datafiles-table-styling)

govuklaurence:
LGTM

PR #222, opened 2017-09-04
Send individual user id

maxf:


app/views/layouts/_analytics.html.erb
```diff
@@ -8,8 +8,10 @@
     })(window,document,'script','https://www.google-analytics.com/analytics.js','ga');
 
     ga('create', trackingId, 'auto');
+    ga('set', 'userId', '<%= session[:beta_user] %>');
     ga('send', 'pageview');
 
+
```
hannako: Laurence might prefer this empty line to not be here :) but in his absence I will approve!

PR #227, opened 2017-09-05
Reinstate tests, fix broken ones

TobyRet:


spec/helpers/datasets_helper_spec.rb
```diff
@@ -19,15 +19,24 @@
...
   it 'returns no_preview if datafile has no preview' do
+    stub_request(:any, 'https://good_data.co.uk').
+      to_return(status: 200)
+
     expect(helper.link_type(DATA_FILES_WITH_START_AND_ENDDATE[1])).to be :no_preview
   end
 
-  fit 'returns preview if datafile has a preview' do
+  it 'returns preview if datafile has a preview' do
+    stub_request(:any, 'https://find-data-beta.herokuapp.com/').
```
maxf: Do we want to have this URL in there? Isn't there a config var that provides it instead of hardcoding it?

TobyRet:
Weird I'm not sure why this PR was closed.

spec/helpers/datasets_helper_spec.rb
TobyRet: The URL is stubbed and we are using dummy data for the test. It could be anything ...

PR #228, opened 2017-09-05
Use dotenv to manage env vars in development

laurentqro:
Just a suggestion here, but I find dotenv to be a practical way to document environment variables and set them locally for development.

https://github.com/bkeepers/dotenv

### DIFF
--- a/.example-env
+++ b/.example-env
@@ -1,0 +1,3 @@
+PUBLISH_URL="https://username:password@publish-data-beta.herokuapp.com"
+SECRET_KEY_BASE="secret" # random string for development
+ES_HOST="127.0.0.1:9200"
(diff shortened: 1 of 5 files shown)


.example-env
```diff
@@ -0,0 +1,2 @@
+PUBLISH_URL="https://username:password@publish-data-beta.herokuapp.com"
```
maxf: I know it's just an example, but I think we should have a fake URL in there.

laurentqro:
@maxf  going to add `.env` to `.gitignore` but it was already there https://github.com/datagovuk/find_data_beta/blob/master/.gitignore#L25. Probably copied from another project.

PR #240, opened 2017-09-07
Jess/content changes consent page

hannako:
Final copy and style changes for use of data page

### DIFF
--- a/app/assets/stylesheets/components.scss
+++ b/app/assets/stylesheets/components.scss
@@ -114,6 +114,9 @@
 
 details {
   margin-bottom: 30px;
+  p:last-child {
+    padding-top: 20px;
+  }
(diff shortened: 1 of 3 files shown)


app/assets/stylesheets/components.scss
```diff
@@ -114,6 +114,9 @@
 
 details {
   margin-bottom: 30px;
+  p:last-child {
```
maxf: Just a small comment: :last-child isn't supported by IE8. It's ok as using it here doesn't break the rule that the application is "functional" in IE8, as mandated by the Service Manual as of this writing: https://www.gov.uk/service-manual/technology/designing-for-different-browsers-and-devices#browsers-to-test-in

PR #244, opened 2017-09-08
Refactor application controller + routes

laurentqro:
### Split into public and private controllers

Setting before_action :authenticate in ApplicationController sets a global default that we have to manually skip every time we write a public controller.

Having controllers for the private areas of the site inherit from a `PrivateAreaController` allows us to avoid having to skip authentication for the public areas of the app.

### ConsentsController

When a user consents to the terms of usage, he in effect creates a new consent. A `ConsentsController#new` allows us to express this in a RESTful way, and clean up the `HomeController`.

### Clean up `authenticate`

* extract local variables to private helper methods
* enforce same level of abstraction

### DIFF
--- a/app/controllers/home_controller.rb
+++ b/app/controllers/home_controller.rb
@@ -1,14 +1,2 @@
-class HomeController < ApplicationController
-  skip_before_action :authenticate, only: [:consent, :confirm_consent]
-
-  def index
-  end
-
-  def consent
-  end
-
-  def confirm_consent
-    session[:consent] = true
-    redirect_to root_path
-  end
+class HomeController < PublicAreaController
 end
--- a/app/controllers/home_controller.rb
+++ b/app/controllers/home_controller.rb
@@ -1,2 +1,2 @@
-class HomeController < PublicAreaController
+class HomeController < LoggedAreaController
 end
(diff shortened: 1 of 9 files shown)


app/controllers/home_controller.rb
```diff
@@ -1,14 +1,2 @@
-class HomeController < ApplicationController
-  skip_before_action :authenticate, only: [:consent, :confirm_consent]
-
-  def index
```
laurentqro: we are following all Rails conventions, so those empty actions are not needed.

app/controllers/logged_area_controller.rb
```diff
@@ -0,0 +1,31 @@
+require 'private_beta_user'
+
+class LoggedAreaController < ApplicationController
+  before_action :authenticate
+end
+
+def authenticate
+  return if Rails.env.development? || Rails.env.test?
```
laurentqro: If it is part of the specification of the app that the user must be logged in, then I'd suggest writing a test helper that signs a user in, and then use that helper in all test that require it.

config/routes.rb
```diff
@@ -1,14 +1,17 @@
 Rails.application.routes.draw do
-  match "/404", :to => "errors#not_found", :via => :all
-  match "/500", :to => "errors#internal_server_error", :via => :all
+  root to: 'home#index'
```
laurentqro: minor. Just more idiomatic to have this at the top.

PR #267, opened 2017-09-13
Set up javascript testing

TobyRet:
Welcome feedback on this!

Have added selenium to allow for javascript testing. It's configured to use a headless browser.
A couple of tests have been added for location and publisher filters. This was a mission as it required manipulating Twitter Typeahead. If I've committed any serious violations please let me know!


Gemfile
```diff
@@ -52,9 +52,11 @@ end
 
 
 group :test do
-  gem "simplecov"
-  gem "codeclimate-test-reporter", "~> 1.0.0"
-  gem "capybara"
+  gem 'simplecov'
+  gem 'codeclimate-test-reporter', '~> 1.0.0'
+  gem 'capybara'
+  gem 'capybara-selenium'
```
TobyRet: Should explicit versioning be used - not sure if we have a policy in place?

spec/features/search_page_spec.rb
```diff
@@ -63,4 +63,80 @@
...
+    assert_data_set_length_is(2)
+
+    page.driver.execute_script %Q{ $('#location').focus().typeahead('val', 'We') }
+    page.driver.execute_script %Q{ $('#location').focus().typeahead('open') }
+
+    expect(page).to have_selector('.tt-selectable', count: 2)
+    expect(page).to have_css('.tt-selectable', text: 'Wellington')
+    expect(page).to have_css('.tt-selectable', text: 'Welding')
+
+    page.driver.execute_script %Q{ $('#location').focus().typeahead('val', 'Wellington') }
```
TobyRet: I was not able to get capybara to select the item from the dropdown. Had to use Jquery to set typeahead's value manually  ... hack ??

laurentqro: This is tied to our implementation, so might prove brittle. But I am guessing you did this because just filling in the field didn't work?

Gemfile
laurentqro: I like Thoughtbot's approach:

```
* Use a versionless Gemfile declarations for gems that are safe to update often, such as pg, thin, and debugger.

* Use an exact version in the Gemfile for fragile gems, such as Rails.
```

https://github.com/thoughtbot/guides/tree/master/best-practices#bundler

spec/features/search_page_spec.rb
laurentqro: Might be a case of Capybara needing to wait until the typeahead value shows up?

.travis.yml
```diff
@@ -11,6 +11,10 @@ rvm:
 services:
   - postgresql
   - elasticsearch
+addons:
+  chrome: beta
+before_install:
+  - "/Applications/Google\\ Chrome.app/Contents/MacOS/Google\\ Chrome --headless --disable-gpu --remote-debugging-port=9222 http://localhost &"
 before_script:
```
maxf: Can you have an environment variable in there to avoid the MacOS-specific stuff?

spec/features/search_page_spec.rb
TobyRet: I tried for ages to try and get Capybara to do this without executing javascript but had no luck. I was reading that apparently trying to get Capybara to play nicely with third party javascript can be a bit hit and miss - it's often the driver that the unknown quantity. Then again, I am a Capybara noob and might not be doing things correctly. Would you be up for having a go?

TobyRet: Ok - will add a sleep and see if this works

Gemfile
TobyRet: OK good advice thanks

.travis.yml
TobyRet: Ok will look into this

TobyRet:
@maxf - the mac-specific stuff has been removed from travis.yml as it is not required. Configuration for using headless browser is in the rails_helper which is not operating system specific

spec/features/search_page_spec.rb
TobyRet: Hey @laurentqro - no luck. Capybara fills in the field ok but for some reason this does not seem to fire an event (which ultimately gets picked up by Typeahead)

PR #271, opened 2017-09-13
Refactor datafile-related logic

laurentqro:
The most significant change in this PR is to parse the datafile hashes into `Dataset` objects. That makes it much easier to manipulate in that we avoid passing around primitives everywhere.


laurentqro:
As far as I know, there is no golden rule, but I consider Rails view helpers to be the place to put complex bits of view code, rather than model-related computations. You could argue that our `Datafile` model here act as a presenter of sorts.

e.g. In a Rails view helper, I'd expect helper methods like:

```
def add_person_to_project_check_box(person, company)
  check_box_tag("people_ids[]", person.id, false, { :class => "company_#{company.id}_person" })
end
```

app/helpers/datasets_helper.rb
```diff
@@ -51,68 +51,28 @@ def expected_location_class_for(dataset)
...
-  end
-
-  def datafiles_with_start_dates(datafiles)
-    datafiles.select { |file| file['start_date'].present? }
-  end
-
   def datafile_next_updated(dataset)
     freq = dataset.frequency
-    last = most_recent_date(dataset.datafiles)
+    last = Time.parse(most_recent_date(dataset.datafiles))
```
hannako: Calling `.advance` on `last` on the next line throws an error if the `most_recent_date` is parsed.

app/models/dataset.rb
```diff
@@ -21,6 +21,18 @@ class Dataset
                 :_index, :_type, :_id, :_score, :_source,
                 :_version
 
+  def datafiles
+    @datafiles.map { |file| Datafile.new(file) }
```
hannako: Can't you lose the @ as there is an attr_accessor for datafiles?

hannako: Also I think this means on the `views/datasets/show` page we need to have:

 
```
<% unless @dataset.datafiles%>
    <%= render partial: "timeseries_data", locals: { datafiles: @timeseries_datafiles } %>
  <% else %>
    <%= render partial: "non_timeseries_data", locals: { datafiles: @non_timeseries_datafiles } %>
  <% end %>
```


Because this current branch has this bug:


screen shot 2017-09-14 at 13 30 23

laurentqro: `attr_accessor` gives me this getter method for free:

```
def datafiles
  @datafiles
end
```

I am overwriting this getter method so it returns an array of objects.

app/helpers/datasets_helper.rb
laurentqro: Which is what the code was already doing. Good spot! I'll ship a fix.

laurentqro: Hm, actually just tried in the Rails console, it works.

```
irb(main):004:0> Time.parse("1/1/15").advance(years: 1)
=> 2002-01-15 00:00:00 +0000
```

laurentqro: I guess it's because

```
irb(main):005:0> Time.parse("1/1/15").class
=> Time
```

Never used `advance` before, cool little method.

PR #273, opened 2017-09-15
Fix show-more toggle for datafiles

TobyRet:


laurentqro:
Works well. Only small niggle: mouse pointer (finger) does not appear when I hover the "show more/less" link.

app/assets/stylesheets/components.scss
```diff
@@ -225,6 +226,11 @@
   &__download {
     margin-top: 1.5em;
   }
+  .show-toggle {
```
maxf: This should use the BEM convention now.

app/views/datasets/_datafile_table.html.erb
```diff
@@ -6,39 +6,45 @@
...
-          <%= link_to (datafile.name ? datafile.name : 'Data Link'),
-              datafile.url,
-              :data => {:'ga-event' => "datafile-download"}
-          %>
-        </td>
-        <% if datafile.format.blank? %>
-          <td class="dgu-unavailable">N/A</td>
-        <% else %>
-          <td><%= datafile.format.upcase %>
+  <% datafiles.each do |datafile| %>
```
maxf: can you indent under `<tbody>` ?

app/views/datasets/_datafile_table.html.erb
```diff
@@ -6,39 +6,45 @@
...
-        </tr>
-      <% end %>
+      </td>
+    </tr>
+  <% end %>
   </tbody>
 </table>
+
+<% if datafiles.length > 5 %>
+  <p class="show-toggle">
```
maxf: this shoud be a button, for accessibility reason. Hence you can reuse the button.secondary class from govuk-overrides.scss

app/assets/stylesheets/components.scss
TobyRet: This is overriding a gov uk class. 
Have changed to 'dgu-datalinks__toggle'

app/views/datasets/_datafile_table.html.erb
TobyRet: Cool - done

app/assets/stylesheets/components.scss
TobyRet: No longer required as we are now using 'button secondary' as requested below

TobyRet:
@laurentqro - hey mate thanks for the comment - point has been fixed

laurentqro:
🙌🏻

app/views/datasets/_datafile_table.html.erb
```diff
@@ -7,38 +7,42 @@
...
-          <%= link_to (datafile.name ? datafile.name : 'Data Link'),
-              datafile.url,
-              :data => {:'ga-event' => "datafile-download"}
-          %>
-        </td>
-        <% if datafile.format.blank? %>
-          <td class="dgu-unavailable">N/A</td>
-        <% else %>
-          <td><%= datafile.format.upcase %>
+    <tr class="<%= show_more?(datafiles, datafile) %> dgu-datafile">
```
maxf: Identation was correct there. Maybe I wasn't clear, sorry: so far we've indented the html inside the <% for/if %> statements (like below). I'm happy to discuss changing that, but we should be consistent

PR #274, opened 2017-09-15
Fix datafiles group by year on dataset page

laurentqro:
@BlancaTortajada When we have both timeseries and non-timeseries datafiles, it looks odd/confusing to group timeseries datafiles under a header (the year) but not group the non-timeseries datafiles.

So here is what this pull request makes it look like for the different cases:

### Timeseries datafiles only

timeseries

### Non timeseries datafiles only

non

### Timeseries + non-timeseries datafiles

group

Also, it's unclear to me in what case we want to show the "Download all" button.

We currently show it when:

* there are only timeseries datafiles
* there are both timeseries and non-timeseries datafiles

We do not show it when we have non timeseries datafiles only.

Not sure if that's the expected behaviour.

Could you please clarify?

### DIFF
--- a/app/views/datasets/_timeseries_data.html.erb
+++ b/app/views/datasets/_timeseries_data.html.erb
@@ -1,20 +1,20 @@
 <ul>
   <% datafiles.group_by(&:start_year).each do |year, datafiles| %>
     <li class="showHide">
-    <div class="year-expand showHide-control">
-      <h3 class="heading-small">
-        <button class="button secondary">
-          <%= year %>
-        </button>
-      </h3>
-      <div>
-        <button class="expand button secondary">+</button>
+      <div class="year-expand showHide-control">
+        <h3 class="heading-small">
+          <button class="button secondary">
+            <%= year %>
+          </button>
+        </h3>
+        <div>
+          <button class="expand button secondary">+</button>
+        </div>
       </div>
-    </div>
-    <div class="year-datasets showHide-content" style="display:none">
-      <%= render partial: 'datafile_table', locals: { datafiles: datafiles } %>
-    </div>
-  <% end %>
+      <div class="year-datasets showHide-content" style="display:none">
+        <%= render partial: 'datafile_table', locals: { datafiles: datafiles } %>
+      </div>
+    <% end %>
   </li>
 </ul>
 
--- a/app/views/datasets/_timeseries_data.html.erb
+++ b/app/views/datasets/_timeseries_data.html.erb
@@ -11,7 +11,7 @@
           <button class="expand button secondary">+</button>
         </div>
       </div>
-      <div class="year-datasets showHide-content" style="display:none">
+      <div class="showHide-content" style="display:none">
         <%= render partial: 'datafile_table', locals: { datafiles: datafiles } %>
       </div>
     <% end %>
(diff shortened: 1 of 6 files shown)


app/views/datasets/_timeseries_data.html.erb
```diff
@@ -1,33 +1,21 @@
...
-              </div>
-          <% end %>
-          </li>
-        </ul>
-        <input class="button dgu-datalinks__download" type="submit" value="Download all">
-      </section>
+<ul>
+  <% datafiles.group_by(&:start_year).each do |year, datafiles| %>
+    <li class="showHide">
+    <div class="year-expand showHide-control">
```
maxf: (Sorry to be pedantic) could you indent under `<li>`?

app/views/datasets/_timeseries_data.html.erb
```diff
@@ -1,33 +1,21 @@
...
+          <%= year %>
+        </button>
+      </h3>
+      <div>
+        <button class="expand button secondary">+</button>
+      </div>
     </div>
-  </div>
-</section>
+    <div class="year-datasets showHide-content" style="display:none">
```
maxf: I don't think that `year-datasets` is used. Feel free to remove, if it doesn't break things.

app/views/datasets/show.html.erb
```diff
@@ -3,8 +3,26 @@
...
+      <div class="column-full">
+        <section class="dgu-datalinks">
+          <h2 class="heading-medium">Data links
+            <% if @timeseries_datafiles.present? %>
+              <!-- accordions for time series datasets -->
+              <span class="showHide-open-all">Open all</span>
+            <% end %>
+          </h2>
+
+          <div class="breaker"></div>
```
maxf: that can also go, I think

app/views/datasets/_timeseries_data.html.erb
laurentqro: Sure thing. Good spot :)

laurentqro:
@maxf Sure, I'll write tests as soon as Blanca has clarified the requirements.

laurentqro:
@maxf added some tests. If requirements change following Blanca's input, I'll update the specs. Should not be too much work.

BlancaTortajada:
Hi @laurentqro,

I agree that when there are other sections with data files grouped by year, the rest of data files without a year should have a heading that groups them, it would look disjointed otherwise. I think we should name this section "Other data files" for now to make it more user friendly and ask Ray to review the copy for that too.

As for "Download all", the expected behaviour is that when the user clicks on "Download all" all the data files in the dataset would be packaged into a zip file and download it. I've had a look at the Trello board, we have a card for this: https://trello.com/c/2WM8y4mq/296-user-can-download-all
We said we'd explore an MVP approach, but as far as I can remember, I don't think we agreed what that approach should be and that there might be perceived security risks from the user's perspective, so probably now is a good time to bring this up again.

Also, please note that the following styling issues:
- The year headings are in NTA Transport and not Helvetica, they should be Helvetica.
- The year headings should not be underlined.
- The plus and minus signs are now being displayed at the same time, but the plus should only be displayed when the section is collapsed and the minus when the section is expanded.

PR #282, opened 2017-09-20
Refactor search controller

TobyRet:
Main change is the creation of a 'Datasets' domain concept which provides a list of publishers and locations

### DIFF
--- a/app/controllers/search_controller.rb
+++ b/app/controllers/search_controller.rb
@@ -1,50 +1,35 @@
 class SearchController < LoggedAreaController
   include QueryBuilder
 
+  before_action :search_for_dataset, only: [:search]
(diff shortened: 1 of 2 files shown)


app/controllers/search_controller.rb
```diff
@@ -1,50 +1,35 @@
...
   end
 
-  def get_locations
-    results = Dataset.search(locations_aggregation_query)
-
-    results.response.aggregations.locations.buckets.map do |bucket|
-      "#{bucket[:key]}"
-    end
+  def make_available_to_js
+    gon.publishers = Datasets.publishers(publishers_aggregation_query)
```
TobyRet: Note that this is making an array of strings available to the javascript in order to populate the publisher and locations autocompletes (can't use ruby objects)

PR #283, opened 2017-09-20
remove uneeded copy

hannako:
BEFORE:

screen shot 2017-09-20 at 11 32 16

AFTER:

screen shot 2017-09-20 at 11 31 51

### DIFF
--- a/app/views/datasets/preview.html.erb
+++ b/app/views/datasets/preview.html.erb
@@ -30,7 +30,7 @@
           </tbody>
         </table>
       <% else %>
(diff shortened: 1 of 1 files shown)


laurentqro:
Does a dataset always have a name (the datafile gets its name from its parent dataset)?

hannako:
@laurentqro The datafile has its own name, it doesn't take it from the parent dataset. On the import, we set the datafile name like [so]( https://github.com/datagovuk/publish_data_beta/blob/2911a8d620ad30e95404752da30477758b3fa379/lib/util/metadata_tools.rb#L62):
```
datafile.name = resource["description"]
datafile.name = "No name specified" if datafile.name.strip() == ""
```

PR #284, opened 2017-09-20
[Spike in progress] Refactor to use id instead of name

laurentqro:


govuklaurence:
Closing but keeping branch in case we want to come back to this

PR #285, opened 2017-09-21
Maxf/display inspire

maxf:


app/views/datasets/_additional_info.html.erb
```diff
@@ -1,9 +1,89 @@
...
-  </p>
-</section>
+<% if @dataset.notes.present? || @dataset.inspire_dataset.present? %>
+
+  <section class="dgu-additional-info">
+    <h2 class="heading-medium">
+      <%= t('.additional_info') %>
+    </h2>
+
+    <% if @dataset.notes.present? %>
```
laurentqro: Checking for the presence of something before displaying is such a common pattern in Rails that it provides us with `presence`, which is exactly equivalent to `object.present? ? object : nil`.

https://apidock.com/rails/Object/presence

maxf: I'm not sure how you'd be using it in a template. Can you give an example?

laurentqro: Just noticed notes are isnide a `<p>` tag, please ignore my comment, it's not relevant here.

laurentqro:
Silly question; what's "inspire"?

PR #286, opened 2017-09-21
Don't track before people have consented

maxf:


app/views/layouts/_analytics.html.erb
```diff
@@ -1,4 +1,4 @@
-<% if Rails.application.config.analytics_tracking_id or Rails.application.config.analytics_test_tracking_id %>
+<% if (Rails.application.config.analytics_tracking_id or Rails.application.config.analytics_test_tracking_id) and not request.url.include?('/use-of-data') %>
```
govuklaurence: `||`, `&&`, `!`

app/views/layouts/_analytics.html.erb
```diff
@@ -1,4 +1,4 @@
-<% if Rails.application.config.analytics_tracking_id or Rails.application.config.analytics_test_tracking_id %>
+<% if (Rails.application.config.analytics_tracking_id || Rails.application.config.analytics_test_tracking_id) && !request.url.include?(new_consent_path) %>
```
TobyRet: Won't this always return true? (Assuming that analytics_tracking_id is always available)

T || F = T

govuklaurence: Nope -- it'll retuurn the left half if it is not nil, else the right half -- but yes it will likely be truthy.

HOWEVER && has a higher precedence than || so it can be false.

PR #287, opened 2017-09-21
Extract Elastic Search queries to their own classes

laurentqro:
My original idea was to wrap all the ElasticSearch stuff into a module of ours, but found myself essentially rewriting the elasticsearch-model gem. It also gives us nice stuff like pagination, so worth relying on it.

Summary:

* Extract ES queries to their own class
* Rely on `Dataset.search(query)` (elasticsearch-model gem) across the codebase
* General DRYing up and reorganising code


app/models/dataset.rb
```diff
@@ -34,33 +54,9 @@ def non_timeseries_datafiles
     datafiles.select(&:non_timeseries?)
   end
 
-  class << self
-    def from_json(raw_json)
-      d = Dataset.new(raw_json.merge(raw_json['_source']))
-      d.json = raw_json
```
TobyRet: @laurentqro - I notice that the 'json' property is no longer there. Might be worth checking with @govuklaurence if this is ok. I've never used it but there may be a reason why it is there.

govuklaurence: It's most likely way outdated code that was left over

govuklaurence:
LGTM, thanks for taking the time to improve this

PR #295, opened 2017-10-04
Preview page improvements

hannako:
This PR relates to [this trello ticket](https://trello.com/c/GsqjYnqE/12-display-line-count-information)

- As per Blanca's request, we no longer need to display "You are previewing x lines out of a total y".
- Also fixes a bug (https://find-data-beta.cloudapps.digital/dataset/public-toilets-st-helens-council-csv) caused by datafiles with a file type of CSV when they are in fact HTML. Now they display the correct preview empty state design.
- I have only come across this bug once, so I've not checked for all other scenarios at this stage.


hannako:
[Display line count information](https://trello.com/c/GsqjYnqE/12-display-line-count-information)

PR #307, opened 2017-10-27
Sentry Install - Find

govuklaurence:


config/application.rb
```diff
@@ -16,8 +16,12 @@ class Application < Rails::Application
...
     config.analytics_test_tracking_id = ENV['GA_TEST_TRACKING_ID']
     config.i18n.load_path += Dir[Rails.root.join('config', 'locales', '**','**','*.{rb,yml}')]
-    # Settings in config/environments/* take precedence over those specified here.
-    # Application configuration should go into files in config/initializers
-    # -- all .rb files in that directory are automatically loaded.
+    config.filter_parameters << :password
+    config.filter_parameters << :password_confirmation
+
+    Raven.configure do |config|
+      config.dsn = 'https://b9b9fe3fe0e84d95a2a2b513b7f859a7:e038e55809d64b2ba482e23031d0e21f@sentry.io/236517'
```
maxf: hardcoded?

govuklaurence: I guess the worst someone could do is flood us with errors, but you're right, should move this to a config.

govuklaurence:
Will fix with secrets setup that I've been delaying for a couple days

app/controllers/application_controller.rb
```diff
@@ -1,3 +1,14 @@
 class ApplicationController < ActionController::Base
   protect_from_forgery with: :exception
+  before_action :set_raven_context
+
+  private
+
+  def set_raven_context
+    Raven.user_context(name: session[:beta_user])
```
hannako: What does :beta_user contain? I can't find where in the code we add to that session

govuklaurence: https://github.com/datagovuk/find_data_beta/blob/master/app/controllers/logged_area_controller.rb#L21

PR #314, opened 2017-11-21
Show dataset contact details

laurentqro:
Related to https://github.com/datagovuk/publish_data_beta/pull/441.

### DIFF
--- a/app/models/dataset.rb
+++ b/app/models/dataset.rb
@@ -4,12 +4,16 @@ class Dataset
 
   attr_accessor :name, :title, :summary, :description,
                 :location1, :location2, :location3,
+                :foi_name, :foi_email, :foi_phone, :foi_web,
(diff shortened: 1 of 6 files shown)


app/models/dataset.rb
```diff
@@ -58,5 +62,9 @@ def self.map_keys(buckets)
     buckets.map { |bucket| bucket['key'] }
   end
 
+  def organisation=(organisation)
+    @organisation = Organisation.new(organisation)
```
hannako: Not sure I follow how you pass in `organisation` to the initialiser here, but on the Organisation "model" it's not expecting an argument...

laurentqro: I am including the ActiveModel::Model module into my class so I can initialize the model with a hash of attributes. More info: http://guides.rubyonrails.org/active_model_basics.html#model

PR #322, opened 2017-12-06
MVP zendesk feedback flow

hannako:
- this PR implements the MVP feedback flow, [ticket](https://trello.com/c/uuPZ7d7I/72-mvp-support-process-for-find)
- Zendesk does not provide a sandbox environment for testing. I've created a dummy zendesk client which is used in staging and development environments. Testing **does not** use the dummy client, but is stubbed by webmock and uses a fake end point.


app/controllers/support_controller.rb
```diff
@@ -0,0 +1,19 @@
+class SupportController < LoggedAreaController
+  skip_before_action :verify_authenticity_token
```
hannako: I spent some time trying to avoid skipping this verification, but I wasn't able to. Without it, the error is that the token is invalid. Poss caused by caching?

config/initializers/zendesk.rb
```diff
@@ -0,0 +1,59 @@
...
+  end_point = zendesk_config_from_vcap['end_point']
+end
+if Rails.env.test?
+  username = ZENDESK_CONFIG['username']
+  password = ZENDESK_CONFIG['password']
+  api_key = ZENDESK_CONFIG['api_key']
+  end_point = ZENDESK_CONFIG['end_point']
+end
+
+GDS_ZENDESK_CLIENT =
```
govuklaurence: confusing to have this over multiple lines, maybe refactor to method and have the assignment to constant be the result of that method

config/zendesk.yml
```diff
@@ -0,0 +1,10 @@
+test:
+  username: 'test-user@mail.com'
+  password: 'password123'
+  api_key: '123abc'
+  end_point: 'https://test_zendesk_url.com/api/v2'
+
+production:
+  vcap_services: '<%= ENV["VCAP_SERVICES"] %>'
+
+#In both staging and development environments zendesk ticket sending is not enabled
```
govuklaurence: space after #

app/controllers/support_controller.rb
```diff
@@ -0,0 +1,18 @@
+class SupportController < LoggedAreaController
```
laurentqro: The support page is a static page that will sit at the root of the site (`/support`), so I think it should be a controller action in here: https://github.com/alphagov/datagovuk_find/blob/master/app/controllers/pages_controller.rb.

laurentqro: The resource we are dealing with here is a `Ticket`, so it makes sense to have a `TicketsController`.

laurentqro: Interesting article related to this topic: http://jeromedalbert.com/how-dhh-organizes-his-rails-controllers/

hannako: Thanks @laurentqro 👍

config/locales/views/layouts/en.yml
```diff
@@ -7,5 +7,5 @@ en:
     proposition_header:
       menu: "Menu"
     application:
-      page_title: "Find government data"
-      global_header_text: "Find government data"
+      page_title: "Data.gov.uk - Find Data"
```
maxf: Is this the official title now?

hannako: That was a change implemented on master by Toby. I will check in with Gurinder and update if necessary on a different PR

hannako:
@maxf I removed the travis commits and squashed some of the others. Is that enough, or would you like me to squash more?

app/models/zendesk_ticket.rb
```diff
@@ -0,0 +1,46 @@
+class ZendeskTicket
+  include ActiveModel::Validations
+  include ActiveModel::Conversion
+  extend ActiveModel::Naming
+
+  attr_accessor :name, :email, :content, :support
+
+  validates_format_of :email, { with: /@/, message: 'Enter a valid email address'}
```
govuklaurence: would check that the @ isn't at the start or the end

config/initializers/zendesk.rb
```diff
@@ -0,0 +1,68 @@
...
+        config["username"] = elem['credentials']['ZENDESK_USERNAME']
+        config["password"] = elem['credentials']['ZENDESK_PASSWORD']
+        config["api_key"] = elem['credentials']['ZENDESK_API_KEY']
+        config["end_point"] = elem['credentials']['ZENDESK_END_POINT']
+      end
+    end
+  rescue => e
+    Rails.logger.fatal 'Failed to extract zendesk creds from VCAP_SERVICES. Exiting'
+    Rails.logger.fatal zendesk_config['vcap_services']
+    Rails.logger.fatal e
```
govuklaurence: would just reraise the error here instead of log and exit

app/views/pages/support.html.erb
```diff
@@ -1,7 +1,46 @@
...
+      </p>
+
+      <p>
+        For questions about a dataset, you can contact the publisher directly if they’ve provided contact details on the dataset page.
+      </p>
+      <p>
+        To get answers about <%= link_to "GOV.UK", "https://www.gov.uk/" %> you can check the GOV.UK help pages or <%= link_to "contact", "https://www.gov.uk/contact/govuk" %> the team.
+      </p>
+      
+      <form action="tickets/new" method="get">
```
maxf: This form is missing a csrf token. I think there's an automatic way to do that in rails.

app/views/tickets/new.html.erb
```diff
@@ -0,0 +1,73 @@
...
+
+    <h1 class='heading-medium'>
+      <% if @support_queue == 'data'%>
+        Send a data request
+      <% else %>
+        Report a problem or give feedback
+      <% end %>
+    </h1>
+
+      <%= form_for @ticket, url: { action: "create", controller: "tickets" } do |f| %>
```
maxf: Maybe that's the syntax that  inserts the token automatically

laurentqro:
@hannako could please you make sure the vowels are displayed in green and the consonants in pink? :trollface:

PR #328, opened 2017-12-13
Unpublished datasets

laurentqro:
The Legacy concept of "unpublished" is where datasets have published metadata but no data files, and deliberately so.

This PR makes sure we:

* label unpublished datasets as `Not released` on both the search results page and the dataset page
* shows different instructions to contact the publisher, or DGU, based on the presence of contact details

Part of card https://trello.com/c/Ykb4YEZO/166-display-unpublished-datasets
Design: https://projects.invisionapp.com/d/main#/console/11687113/268789175/preview

### DIFF
--- a/app/assets/stylesheets/components.scss
+++ b/app/assets/stylesheets/components.scss
@@ -368,9 +368,3 @@ span.showHide-open-all {
     }
   }
 }
-
-
-// ==== not a component ========================================================
-.dgu-unavailable {
-  color: $secondary-text-colour;
-}
--- a/app/assets/stylesheets/dgu.scss
+++ b/app/assets/stylesheets/dgu.scss
@@ -1,0 +1,8 @@
+.dgu-secondary-text {
+  color: $secondary-text-colour;
+}
+
+.highlight {
+  background: $grey-3;
+  padding: em(2) em(5) em(2) em(5);
+}
--- a/app/assets/stylesheets/dgu.scss
+++ b/app/assets/stylesheets/dgu.scss
@@ -2,7 +2,7 @@
   color: $secondary-text-colour;
 }
 
-.highlight {
+.dgu-highlight {
   background: $grey-3;
   padding: em(2) em(5) em(2) em(5);
 }
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -15,3 +15,4 @@
 @import 'govuk';
 @import 'govuk-overrides';
 @import 'components';
+@import 'dgu';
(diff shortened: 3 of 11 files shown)


Gemfile
```diff
@@ -9,7 +9,7 @@ gem 'jbuilder', '~> 2.5'
 gem 'devise', '~> 4.3'
 gem 'devise_invitable', '~> 1.7.2'
 gem 'activeadmin', '~> 1'
-gem 'govuk_elements_rails'
+gem 'govuk_elements_rails', '~> 3.1'
```
maxf: Could you separate that out to a different commit to explain the change?

app/assets/stylesheets/dgu.scss
```diff
@@ -0,0 +1,8 @@
+.dgu-secondary-text {
+  color: $secondary-text-colour;
+}
+
+.highlight {
```
maxf: Please use our naming convention.

app/assets/stylesheets/components.scss
```diff
@@ -368,9 +368,3 @@ span.showHide-open-all {
     }
   }
 }
-
-
-// ==== not a component ========================================================
-.dgu-unavailable {
```
maxf: Why removed?

laurentqro: We were using `dgu-unavailable` for things that were not ... `unavailable`, so I renamed this to the more general `dgu-secondary-text`.

Moved it to the `dgu.scss` file because it's not a component.

Gemfile
laurentqro: I like to limit gems to non-breaking subsets of versions with the use of `~>` to avoid anyone running `bundle update` and making breaking changes in our app. Ideally all our gems would do that.

http://bundler.io/v1.16/guides/updating_gems.html

app/assets/stylesheets/dgu.scss
laurentqro: `dgu-highlight`?

Gemfile
maxf: yes. Can you put it in a different commit though, explaining the above. Ideally doing it for all gems in one go? Just so we have a cleaner history of code changes.

app/assets/stylesheets/components.scss
maxf: ok. We'll have to make it a component and try and use more semantic naming. But I'll take care of that later.

app/assets/stylesheets/dgu.scss
maxf: yes

Gemfile
laurentqro: ok @maxf I'll do this in a separate PR.

laurentqro: I removed the commit with the Gemfile changes.

PR #334, opened 2017-12-19
Force UTF-8 encoding for all preview responses

maxf:


laurentqro:
`force_encoding` changes the rules for interpreting the data contained in the previewed file, but it doesn't change the data.

We could get away with this, as long as `ASCII-8BIT` and `UTF-8` mean the same thing. But it will blow up if it doesn't, along the lines of `invalid byte sequence in UTF-8`.

Instead, we might consider using `encode`.

Very good article about this: http://graysoftinc.com/character-encodings/ruby-19s-string.

laurentqro:
this one's good too https://www.justinweiss.com/articles/3-steps-to-fix-encoding-problems-in-ruby/.

PR #342, opened 2018-01-03
Use require instead of sass @import directive

laurentqro:
Not entirely sure why but `@import 'accessible-autocomplete.min.css'` literally gets processed into `@import url(accessible-autocomplete.min.css)`, which points to a non-existent file (the Rails asset pipeline processes all the CSS into a single file).

The good old `*/ = require` should fix this.


laurentqro:
You're too fast for me, I was in the process of writing the description. ;-) I'll make sure to write my descriptions before I open the PR in the future :)

PR #345, opened 2018-01-04
Redirect legacy search URLs to beta search results page

laurentqro:
This PR makes sure we redirect any legacy search URL to the search results page on Find Beta, in a way that includes the original search query and filters.

https://trello.com/c/NQa9b2KZ/225-find-redirect-searches-on-legacy-dgu-to-search-results-page-on-beta

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -14,6 +14,7 @@ Rails.application.routes.draw do
 
   get 'search/', to: 'search#search'
   get 'search/tips', to: 'search#tips'
+  get 'data/search', to: 'legacy_search#redirect'
 
   get 'dataset/:legacy_name', to: 'legacy_datasets#redirect'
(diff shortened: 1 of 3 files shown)


config/routes.rb
```diff
@@ -14,6 +14,7 @@
 
   get 'search/', to: 'search#search'
   get 'search/tips', to: 'search#tips'
+  get 'data/search', to: 'legacy_search#redirect'
```
TobyRet: @laurentqro what do you think about splitting the routes.rb so its clearer which ones are legacy?
https://blog.arkency.com/2015/02/how-to-split-routes-dot-rb-into-smaller-parts/

laurentqro: Good point. 👍 

Splitting into a separate file might be premature, but we could start namespacing legacy stuff, that would already make things much clearer. Can do that in a separate PR.

PR #352, opened 2018-01-15
Move attribute where it's supposed to be

maxf:


maxf:
@hannako : having it on the tr didn't pass Google's validation tool. I enquired and was told it should go on the tr. I don't think that's very logical though. The reason why it's distribution and not datafile is in the taxonomy for Dataset. See http://schema.org/Dataset

PR #354, opened 2018-01-17
Fix autocomplete on datafile format filter

laurentqro:
### Bug

On arriving on the search page, with no filters selected, the datafile format filter already has one selected (the first one in the list).

### Fix

The accessible autocomplete automatically selects the first value in the list of possible values that is given to it. In this case, the list of available format filters is given to us by an elastic search aggregation query. For the filter to work with the expected behaviour, the first item in the list MUST be an empty string.

However, because the results of the aggregation are based on the datafile formats present in the ES index, this list will sometimes be in the correct format (i.e. `["", "foo", "bar"]`), sometimes not, depending on whether a datafile in the index happens to have a format value set as `""`.

To avoid this, this PR makes sure we always prepend an empty string to our list of formats.

### DIFF
--- a/app/helpers/search_helper.rb
+++ b/app/helpers/search_helper.rb
@@ -9,7 +9,7 @@ module SearchHelper
   end
 
   def datafile_formats_for_select
-    Dataset.datafile_formats
+    Dataset.datafile_formats.unshift("").uniq
   end
 
   def selected_format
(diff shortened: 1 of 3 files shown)


app/helpers/search_helper.rb
```diff
@@ -9,7 +9,7 @@ def format(timestamp)
   end
 
   def datafile_formats_for_select
-    Dataset.datafile_formats
+    Dataset.datafile_formats.unshift("").uniq
```
maxf: single quotes please. https://standardjs.com/rules.html

PR #361, opened 2018-01-18
Format filter fix

maxf:
This fixes a few front-end issues:
- removes unnecessary code and makes a few things consistent with the rest of the code
- shows the arrow on the dropdown, as reported at https://trello.com/c/46ssQEzV/119-bug-find-downward-arrow-should-appear-in-dropdown-menu-of-the-datafile-format-filter

We still use accessible-autocomplete because of the styling of the dropdown, but it's a bit on the heavy side that we have to use both tt-typeahead and accessible-autocomplete on the same page.


app/helpers/search_helper.rb
```diff
@@ -9,7 +9,7 @@ def format(timestamp)
   end
 
   def datafile_formats_for_select
-    Dataset.datafile_formats.unshift('').uniq
+    Dataset.datafile_formats.map(&:downcase).uniq.sort.unshift('')
```
laurentqro: ElasticSearch natively sorts that by number of results, but ok why not :)

laurentqro: Need to update the description of the spec (`it 'returns a list of unique datafile formats ordered by datafile count (elasticsearch default)`)

app/views/search/_filters.html.erb
```diff
@@ -18,7 +18,7 @@
 
 <div class="form-group">
   <%= label_tag 'filters[format]', t('.filter_by_format'), class: 'form-label' %>
-  <%= select_tag('filters[format]', options_for_select(datafile_formats_for_select, selected_format), id: 'js_format-filter', class: 'form-control dgu-filters__filter') %>
+  <%= select_tag('filters[format]', options_for_select(datafile_formats_for_select, selected_format), id: 'format', class: 'form-control dgu-filters__filter') %>
 </div>
```
laurentqro: the reason why I prefixed this id name with `js_` is to indicate that it is being used by javascript. I am afraid someone will remove this in the future, thinking it's only cosmetic.

app/helpers/search_helper.rb
maxf: - It wasn't sorting anything before I added the sort.
- will do, thanks.

app/views/search/_filters.html.erb
maxf: I don't think it's a good idea to give an id to something based on its current use. If someone removes it without checking if it's being used, they're asking for it...

PR #362, opened 2018-01-18
Refactor models on Find

hannako:
This PR relates to https://trello.com/c/S2dYl1vy/97-bug-documents-are-shown-in-datafiles-sections

- Changes have been made to our data models on Publish. We now have Dataset.links with type datafile or type doc. See [Publish PR #493](https://github.com/alphagov/datagovuk_publish/pull/493)
- In addition we are now storing Datafiles and Docs as separate arrays on elastic search
- As a result changes were required on Find. Namely the creation of a new model: Doc, and a refactoring of the existing model Datafile.

### DIFF
--- a/app/models/dataset.rb
+++ b/app/models/dataset.rb
@@ -8,14 +8,14 @@ class Dataset
                 :contact_name, :contact_email, :contact_phone,
                 :licence, :licence_other, :frequency,
                 :published_date, :last_updated_at, :created_at,
-                :harvested, :uuid, :datafiles,
+                :harvested, :uuid,
                 :inspire_dataset, :json, :notes,
                 :contact_name, :contact_email, :contact_phone,
                 :foi_name, :foi_email, :foi_phone, :foi_web,
                 :_index, :_type, :_id, :_score, :_source,
                 :_version
 
-  attr_reader :organisation
+  attr_reader :organisation,:docs, :datafiles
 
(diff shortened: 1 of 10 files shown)


app/models/dataset.rb
```diff
@@ -62,8 +62,16 @@ def self.datafile_formats
...
+  def docs=(docs)
+    @docs = docs.map { |file| Doc.new(file) }
+  end
+
+  def datafiles=(datafiles)
+    @datafiles = datafiles.map { |file| Datafile.new(file)}
+  end
+
+  def links
+    @docs + @datafiles
```
laurentqro: I'd call the reader methods here rather than the instance vars.

app/views/datasets/show.html.erb
```diff
@@ -41,7 +41,7 @@
 
   <%= render "additional_info" %>
 
-  <% unless no_documents?(@dataset.datafiles) %>
+  <% unless @dataset.docs.empty? %>
```
laurentqro: We'll need to make sure `@dataset.docs` returns `[]` here. :)

https://sentry.io/government-digital-service/find/issues/437118602/

hannako: It does return [ ] this error was because the ES mappings hadn't been updated on Publish. I opened a PR to fix. https://github.com/alphagov/datagovuk_publish/pull/500

PR #364, opened 2018-01-22
Fix undefined method #empty? for nil:NilClass

laurentqro:
Fixes https://sentry.io/government-digital-service/find/issues/437118602/

Coercing `@datafiles` and `@docs` into an array ensures the reader methods always returns an array, as can reasonably be expected.


laurentqro:
thanks @maxf

PR #380, opened 2018-01-30
Display welcome message on all pages

hannako:
This PR relates to: https://trello.com/c/wyFRwDcF/17-message-on-homepage-to-explain-change-to-users

- Shows a single blue banner on all pages containing a link to provide feedback and an explanation about the new dgu service. Once dismissed, the standard beta banner is displayed.

Bug fixes:
- Styling on static pages
- Reinstates beta banner on all pages (was missing from landing and static pages)

### DIFF
--- a/app/assets/stylesheets/components.scss
+++ b/app/assets/stylesheets/components.scss
@@ -71,61 +71,52 @@
   a:hover {
     font-weight: bold;
   }
-}
 
-// when beta message is displayed, spacing of adjacent elements must be adjusted____________
-
-// 1. Static pages and home page
-.dgu-beta-message__show {
-  .grid-row.inner {
-    .column-two-thirds {
-      h1.heading-xlarge {
-        margin-top: 40px;
-      }
-      h1.heading-large {
-        margin-top: 40px;
-      }
-    }
-  }
-}
-
-// 2. Dataset show page
-.dgu-beta-message__show {
-  div {
-    .meta-data {
-      .grid-row {
-        .column-two-thirds {
-          h1.heading-large {
-            margin-top: 40px;
-          }
+  // when beta message is displayed, spacing of adjacent elements must be adjusted____________
+  &--show {
+    // 1. Static pages and home page
+    .grid-row.inner {
+      .column-two-thirds {
+        h1.heading-xlarge {
+          margin-top: 40px;
         }
-      }
-    }
-  }
-}
-
-// 3. Search results page
-.dgu-beta-message__show {
-  form {
-    .dgu-split-background__top {
-      .dgu-split-background__top-inner {
         h1.heading-large {
           margin-top: 40px;
         }
+      }
+    }
+    // 2. Dataset show page
+    div {
+      .meta-data {
         .grid-row {
-          .dgu-beta_message {
-            margin-top: 40px;
+          .column-two-thirds {
+            h1.heading-large {
+              margin-top: 40px;
+            }
           }
         }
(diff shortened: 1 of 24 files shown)


app/assets/stylesheets/components.scss
```diff
@@ -34,6 +34,76 @@
...
+  }
+  a:hover {
+    font-weight: bold;
+  }
+}
+
+// when beta message is displayed, spacing of adjacent elements must be adjusted____________
+
+// 1. Static pages and home page
+.dgu-beta-message__show {
```
maxf: you can group everything under the top level rule on line 38, using &__show {...}

maxf: According to BEM naming conventions (http://getbem.com/naming/) the name should be
`.dgu-beta-message--show` as `show` is a modifier for the class

app/assets/stylesheets/components.scss
```diff
@@ -34,6 +34,76 @@
...
+      }
+    }
+  }
+}
+
+// 3. Search results page
+.dgu-beta-message__show {
+  form {
+    .dgu-split-background__top {
+      .dgu-split-background__top-inner {
```
maxf: You can use `&__top-inner` to avoid duplication

app/controllers/application_controller.rb
```diff
@@ -11,4 +12,13 @@ def set_raven_context
                         environment: Rails.env,
                         app: ENV['VCAP_APPLICATION'])
   end
+
+  def toggle_beta_message
+    flash[:beta_message] =
+      beta_message_unseen? ? 'show' : 'hide'
```
maxf: it seems that 'hide' is not needed as a value and could be empty

TobyRet:
I'll make the changes in Jess's absence

app/controllers/application_controller.rb
TobyRet: flash[:beta_message] needs to be assigned something regardless of whether beta_message_unseen? return true or false

app/controllers/search_controller.rb
```diff
@@ -1,5 +1,6 @@
 class SearchController < LoggedAreaController
   before_action :search_for_dataset, only: [:search]
+  before_action :toggle_beta_message
```
laurentqro: already present in `ApplicationController` and `LoggedAreaController` inherits from it, so this `before_action` is not need I think.

andrewgarner:
beta

Is the bold styling required when you hover over the links? It's a bit jarring having it wrap each time.

TobyRet:
Yes I thought that should start on a new line @hannako ?

andrewgarner:
Even with a line break we still might see the same behaviour depending on device width.

maxf:
I'll check with the front-end community. That's appeared before.

maxf:
@hannako . Alternating bold / normal is an anti-pattern because of this problem, but also because it makes some users (specifically users of assistive tech like screen magnification) confused because things change on the screen unexpectedly and with no specific reason.

app/controllers/search_controller.rb
hannako: 👍

andrewgarner:
There is a path through this that doesn't quite work when the application is in private beta.

You are shown the message on the "How we’ll use your data" which you see if you haven't opted into the beta. The "get in touch" link is private to the beta, so clicking on it just takes you back to the page you're already on. There is no way to get in touch without joining the beta.

PR #391, opened 2018-02-12
Use Rails' custom configuration support to configure Elasticsearch

andrewgarner:
Rails has built-in support for custom configuration. This takes care of parsing configuration files, and selecting the correct configuration values for the current environment.

This changes the way we configure Elasticsearch to use this strategy instead of rolling our own implementation.

I unearthed this because I couldn't run the application without a `RAILS_ENV` variable set which was required in line 7 of `config/initializers/elasticsearch.rb`.

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -35,6 +35,8 @@ module FindDataBeta
     config.filter_parameters << :password
     config.filter_parameters << :password_confirmation
 
+    config.elasticsearch = config_for(:elasticsearch)
+
     Raven.configure do |config|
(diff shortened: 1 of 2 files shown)


config/application.rb
```diff
@@ -35,6 +35,8 @@ class Application < Rails::Application
     config.filter_parameters << :password
     config.filter_parameters << :password_confirmation
 
+    config.elasticsearch = config_for(:elasticsearch)
```
TobyRet: Nice - way simpler

PR #393, opened 2018-02-13
Redirect legacy dataset resource pages

andrewgarner:
Adds a permanent redirect for dataset resource pages to the preview page. For example https://data.gov.uk/dataset/historic-flood-map1/resource/4f661185-1788-4d61-85e3-d843c6fc914a

### DIFF
--- a/spec/requests/legacy/redirect_spec.rb
+++ b/spec/requests/legacy/redirect_spec.rb
@@ -38,6 +38,41 @@ describe 'legacy', :type => :request do
     end
   end
 
+  describe 'datafile resources' do
+    let(:legacy_dataset_name) { 'a-legacy-name' }
+    let(:datafile) { CSV_DATAFILE.with_indifferent_access }
+    let(:dataset) do
+      DatasetBuilder
+        .new
+        .with_legacy_name(legacy_dataset_name)
+        .with_datafiles([datafile])
+        .build
+    end
+
+    context 'when the datafile can not be found' do
+      it 'returns a not found error page' do
+        get "/dataset/#{legacy_dataset_name}/resource/#{datafile[:uuid]}"
+
+        expect(response).to have_http_status(:not_found)
+      end
+    end
+
+    context 'when the datafile exists' do
+      before { index([dataset]) }
+
+      it 'redirects to the datefile preview page' do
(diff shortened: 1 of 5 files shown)


spec/requests/legacy/redirect_spec.rb
```diff
@@ -38,6 +38,63 @@
...
+    let(:dataset) do
+      DatasetBuilder
+        .new
+        .with_legacy_name(legacy_dataset_name)
+        .with_datafiles([datafile])
+        .build
+    end
+
+    context 'when the datafile can not be found' do
+      it '' do
```
maxf: It would help if you added the 'it' part too, so that the human-readable description is complete.

spec/requests/legacy/redirect_spec.rb
```diff
@@ -38,6 +38,63 @@
...
+      end
+    end
+
+    context 'when the datafile can not be previewed' do
+      before do
+        stub_request(:get, datafile[:url]).to_return(status: [404, "Not Found"])
+        index([dataset])
+      end
+
+      it 'redirects to the files part of the dataset page' do
```
maxf: I'm not sure about that redirect. Even if the dataset can't be previewed, the preview page exists and not only says 'can't be previewed' but also offers a download button.

spec/requests/legacy/redirect_spec.rb
andrewgarner: Good spot 😧

spec/requests/legacy/redirect_spec.rb
andrewgarner: Okay, that's a more elegant solution. I'll do that!

spec/requests/legacy/redirect_spec.rb
andrewgarner: Fixed with https://github.com/alphagov/datagovuk_find/pull/393/commits/0c3c7b043a2ad443c6a36fda2750d2aa46369f14

spec/requests/legacy/redirect_spec.rb
andrewgarner: Fixed with https://github.com/alphagov/datagovuk_find/pull/393/commits/5b5639973b9c6b70e7c509c175bb49f79f8b6e74

PR #395, opened 2018-02-13
Improve search filters

hannako:
This PR:

1. Removes filter by location. 
Ref: https://trello.com/c/dyxZGrtq/203-remove-geography-filter-and-geographical-area-field

2. Prevents users from entering random terms in the publisher filter field, as this was leading to unexpected results. Eg entering 'ministry' returned 0 results, as we were doing an exact match and no datasets are published by 'ministry'. Instead of using the Twitter typeahead we are now using the accessible autocomplete - duplicating the behaviour of the other filters on the page. 
Ref: https://trello.com/c/Mq7rUbRs/200-bug-search-0-results-when-incomplete-publisher-entered-in-filter


andrewgarner:
Do we still need this line?

https://github.com/alphagov/datagovuk_find/blob/bc83ef62f5aeaa09d3ec1156f07c47e34be7834f/app/controllers/search_controller.rb#L26

TobyRet:
Also some more stuff to be removed from search controller
before_action :make_available_to_js, only: [:search] (delete method also)

config/locales/views/search/en.yml
```diff
@@ -27,6 +27,7 @@ en:
       filter_by_pub: "Publisher"
```
andrewgarner: We should remove this if we're going with `filter_by_publisher`.

spec/features/search_page_spec.rb
```diff
@@ -65,51 +65,13 @@
...
-    end
-
-    datasets = all('h2 a')
-
-    expect(datasets.length).to be(1)
-    expect(datasets[0]).to have_content 'Wellington Dataset'
-  end
-
-  scenario 'Match publisher query against available publishers', :js => true do
+  scenario 'Match publisher query against available publishers' do
```
andrewgarner: We're applying the `accessibleAutocomplete` JavaScript component to the publisher field but we're not testing its behaviour with a WebDriver that supports JavaScript. It would be good to add some steps that prove accessible autocomplete is integrated correctly.

config/locales/views/search/en.yml
hannako: Oh thank you - good spot!

spec/features/search_page_spec.rb
hannako: I've changed the test - could you let me know if the changes address your comment?  I wasn't 100% sure what you meant by proving it was integrated correctly.

spec/features/search_page_spec.rb
```diff
@@ -89,15 +89,19 @@
 
     assert_data_set_length_is(2)
 
-    select('Toby Corp', from: 'filters[publisher]')
+    find('#publisher')
+    execute_script %Q{ $('#publisher').val('Toby') }
```
andrewgarner: We can drive the elements on the page without having to use JavaScript:

```rb
scenario 'Match publisher query against available publishers', js: true do
  first_dataset = DatasetBuilder.new
                    .with_title('Data About Tonka Trucks')
                    .with_publisher('Tonka Trucks')
                    .build

  second_dataset = DatasetBuilder.new
                     .with_title('Data About Toby')
                     .with_publisher('Toby Corp')
                     .build

  index [first_dataset, second_dataset]

  visit '/search'

  # FIXME - force the page to scroll so the elements are visible
  execute_script "window.location = '#publisher'"

  find_field('publisher').send_keys('Ton', :down, :enter)
  click_button 'Apply filters'

  search_result_titles = all('h2 a').map(&:text)

  expect(search_result_titles).to contain_exactly('Data About Tonka Trucks')
end
```

PR #397, opened 2018-02-14
Datasets with no topics will display 'not added'

TobyRet:
Some datasets do not have a topic assigned. If so then show 'Topic: Not added' in the dataset meta data panel

Trello - https://trello.com/c/Ot65Qefh/80-implement-topics-themes-l

### DIFF
--- a/app/views/datasets/_meta_data.html.erb
+++ b/app/views/datasets/_meta_data.html.erb
@@ -36,9 +36,11 @@
             <span itemprop="name"><%= dataset_location(@dataset) %></span>
           </dd>
 
+          <dt><%= t('.topic') %>:</dt>
           <% if @dataset.topic %>
-            <dt><%= t('.topic') %>:</dt>
-            <dd><%= @dataset.topic['title'] %>
+            <dd><%= @dataset.topic['title'] %></dd>
+          <% else %>
+            <dd><%= t('.topic_not_added') %></dd>
           <% end %>
 
           <dt><%= t('.licence') %>:</dt>
--- a/spec/features/dataset_show_page_spec.rb
+++ b/spec/features/dataset_show_page_spec.rb
@@ -19,8 +19,17 @@ feature 'Dataset page', elasticsearch: true do
       expect(page).to have_content('Geographical area: London Southwark')
     end
 
-    scenario 'Display the topic' do
+    scenario 'Display the topic if there is one' do
       expect(page).to have_content('Topic: Government')
+    end
+
+    scenario 'Do not display the topic if information missing' do
+      dataset = DatasetBuilder.new.build
+      dataset['topic'] = nil
+      
+      index_and_visit(dataset)
+
+      expect(page).to have_content('Topic: Not added')
     end
   end
 
--- a/config/locales/views/datasets/en.yml
+++ b/config/locales/views/datasets/en.yml
@@ -45,6 +45,8 @@ en:
       published_by: "Published by"
       last_updated: "Last updated"
       geographical_area: "Geographical area"
+      topic: "Topic"
+      topic_not_added: "Not added"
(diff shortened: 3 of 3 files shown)


app/views/datasets/_meta_data.html.erb
```diff
@@ -36,9 +36,11 @@
             <span itemprop="name"><%= dataset_location(@dataset) %></span>
           </dd>
 
-          <% if @dataset.topic %>
-            <dt><%= t('.topic') %>:</dt>
-            <dd><%= @dataset.topic['title'] %>
+          <dt><%= t('.topic') %>:</dt>
+          <% if(@dataset.topic) %>
+            <dd><%= @dataset.topic['title'] %></dd>
```
maxf: is there a method that wouldn't crash if the topic didn't have a title (from bad input data). Python has `dataset.topic.get('title', '')` that provides a default

maxf: In other words we need not to show the user a 500 page, unless it's actually impossible to show anything useful.

andrewgarner: `<%= @dataset.topic['title'] %>` will return the default value, in this case `nil` (and won't crash) if the `topic` hash has no `title` key. See https://docs.ruby-lang.org/en/2.0.0/Hash.html#method-i-5B-5D

The equivalent to `get` in Python would be `fetch`. See https://docs.ruby-lang.org/en/2.0.0/Hash.html#method-i-fetch

app/views/datasets/_meta_data.html.erb
```diff
@@ -36,9 +36,11 @@
             <span itemprop="name"><%= dataset_location(@dataset) %></span>
           </dd>
 
-          <% if @dataset.topic %>
-            <dt><%= t('.topic') %>:</dt>
-            <dd><%= @dataset.topic['title'] %>
+          <dt><%= t('.topic') %>:</dt>
+          <% if(@dataset.topic) %>
```
andrewgarner: The parenthesis are not needed here.

spec/features/dataset_show_page_spec.rb
```diff
@@ -10,18 +10,27 @@
 
   feature 'Meta data' do
     before(:each) do
-      dataset = DatasetBuilder.new.build
+      @dataset = DatasetBuilder.new.build
 
-      index_and_visit(dataset)
+      index_and_visit(@dataset)
```
andrewgarner: Does this need to an instance variable? It's not referenced in the new scenario.

TobyRet: Good spot - thanks

app/views/datasets/_meta_data.html.erb
TobyRet: Old habits die hard

app/views/datasets/_meta_data.html.erb
andrewgarner: Very minor, but we're duplicating the `<dd>` tags. You could be bit fancy:

```erb
<dd><%= @dataset&.dig(:title) || t('.topic_not_added') %></dd>
```

spec/features/dataset_show_page_spec.rb
```diff
@@ -12,16 +12,25 @@
     before(:each) do
       dataset = DatasetBuilder.new.build
 
-      index_and_visit(dataset)
+      index_and_visit(@dataset)
```
andrewgarner: This one too. `index_and_visit(dataset)`

app/views/datasets/_meta_data.html.erb
TobyRet: Can't use dig on ActiveRecord instances :-(

PR #398, opened 2018-02-14
Dashboard of what data is present in the index

laurentqro:
screen shot 2018-02-14 at 4 02 34 pm

https://trello.com/c/9T5RNLeN/142-admin-reports-on-find

### DIFF
--- a/app/views/pages/dashboard.html.erb
+++ b/app/views/pages/dashboard.html.erb
@@ -1,0 +1,62 @@
+<% content_for :page_title do %>
+  What's available in Find
+<% end %>
+
+<h1 class="heading-large">
+  What's available in Find right now
+</h1>
+
+<div class="grid-row">
+  <div class="column-one-third">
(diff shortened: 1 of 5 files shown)


app/views/pages/dashboard.html.erb
```diff
@@ -0,0 +1,62 @@
...
+
+<h1 class="heading-large">
+  What's available in Find right now
+</h1>
+
+<div class="grid-row">
+  <div class="column-one-third">
+    <div class="data">
+      <span class="data-item bold-xlarge"><%= number_with_delimiter(@datasets_count) %></span>
+      <span class="data-item bold-xsmall">datasets</span>
```
andrewgarner: Although an edge case, this would be an improvement:

```erb
<span class="data-item bold-xsmall"><%= pluralize(@datasets_count, 'dataset') %></span>
```

PR #399, opened 2018-02-14
Fix layout

laurentqro:
https://trello.com/c/gOPD1a0e/215-the-layout-of-the-static-pages-has-been-broken-s

### Before

screen shot 2018-02-14 at 4 13 59 pm

### After

screen shot 2018-02-14 at 4 12 44 pm

### DIFF
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -28,7 +28,9 @@
 
 <% content_for :content do %>
   <%= render 'layouts/staging_ribbon' if Rails.env.staging? %>
-  <%= yield %>
+  <main role="main" tabindex="-1" id="<%= 'content' unless current_page? root_path %>">
+    <%= yield %>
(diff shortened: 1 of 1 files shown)


app/views/layouts/application.html.erb
```diff
@@ -28,7 +28,9 @@
 
 <% content_for :content do %>
   <%= render 'layouts/staging_ribbon' if Rails.env.staging? %>
-  <%= yield %>
+  <main role="main" tabindex="-1" id="<%= 'content' unless current_page? root_path %>">
```
andrewgarner: What's special about the root path? Should the application layout know about rendering behaviour at a given path?

laurentqro: I don't know. I lazily copied the implementation over from Publish. Shame on me. :)

laurentqro:
fixed in another PR

PR #403, opened 2018-02-19
Phrase based searches

andrewgarner:
This adds support for phrase based searches. A phrase being any part of the search query that is wrapped in quotes.

To achieve this I have introduced the [Parslet] Ruby library. This allows us to use a formal grammar to define the makeup of search queries. The implemented parser defines rules for term and phrase based queries.

For example, a query of 'flood map for planning' would be parsed as a term based query. A query of '"flood map" for planning' would be parsed as a phrase of "flood map" and the terms "for planning".

An alternative approach would be use Elasticsearch's support for [simple query string] however doing so would add support for a wider search syntax that we don't currently have an identified user need for.

[parslet]: http://kschiess.github.io/parslet/
[simple query string]: https://www.elastic.co/guide/en/elasticsearch/reference/6.2/query-dsl-simple-query-string-query.html

See https://trello.com/c/AyV10xt7

### DIFF
--- a/app/services/search/query_parser.rb
+++ b/app/services/search/query_parser.rb
@@ -1,0 +1,21 @@
+module Search
+  class QueryParser < Parslet::Parser
+    rule(:whitespace) { match['\s'].repeat(1) }
+    rule(:whitespace?) { whitespace.maybe }
+
+    rule(:quote) { str('"') >> whitespace? }
+    rule(:quote?) { quote.maybe }
+
+    rule(:term) { match['[^\s"]'].repeat(1).as(:term) >> whitespace? }
+    rule(:term?) { term.maybe }
+    rule(:terms) { term.repeat(1) }
+
+    rule(:phrase) { quote >> terms.as(:phrase) >> quote }
+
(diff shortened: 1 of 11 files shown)


app/services/search/query_parser.rb
```diff
@@ -0,0 +1,21 @@
+module Search
+  class QueryParser < Parslet::Parser
+    rule(:whitespace) { match['\s'].repeat(1) }
+    rule(:whitespace?) { whitespace.maybe }
+
+    rule(:quote) { str('"') >> whitespace? }
+    rule(:quote?) { quote.maybe }
+
```
hannako: should we have both a `:single_quote?` and a `:double_quote?` rule??

hannako: Probably not - google just does the exact match on double quotes doesn't it. Plus then you'd have issue with matching on things like HRMC's.

andrewgarner: I followed [Google's lead](https://support.google.com/websearch/answer/2466433):

> **Search for an exact match**
> Put a word or phrase inside quotes. For example, "tallest building".

PR #410, opened 2018-02-19
Use English analyser when searching

andrewgarner:
This switches the analyser used for search queries to one that undertands the English language.

Doing so means that “tree in birmingham”, “trees in birmingham”, “tree birmingham”, “trees birmingham” are all treated the same.

See https://www.elastic.co/guide/en/elasticsearch/reference/current/analysis-lang-analyzer.html#english-analyzer

### DIFF
--- a/app/services/search/query.rb
+++ b/app/services/search/query.rb
@@ -196,6 +196,7 @@ module Search
       {
         multi_match: {
           query: query,
+          analyzer: 'english',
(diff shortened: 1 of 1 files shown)


laurentqro:
I guess at some point we'll need to write high-level tests to make sure search is well-behaved :)

maxf:
@laurentqro I'm looking into pingdom in order to have smoke tests. But requires higher-tiered subscription.

PR #414, opened 2018-02-20
Remove http authentication and obligatory consent page

laurentqro:
This is to merge just before we launch Find as open beta, not before.

This PR removes:

* removes https://find-data-beta.cloudapps.digital/use-of-data and stops people being redirected to it
* removes http authentication

https://trello.com/c/py5HzPkP/176-remove-initial-optin-page-and-http-auth-from-find-s


andrewgarner:
I think we can remove `PrivateBetaUser`, `new_user.rake`, `config/initializers/wordlist.rb` too.

laurentqro:
yes, good catch. thanks!

lib/private_beta_user.rb
```diff
@@ -1,27 +0,0 @@
-module PrivateBetaUser
```
andrewgarner: This is referenced by `new_user.rake` so you'll need to also remove that!

PR #429, opened 2018-02-23
Make new site-changes page

nicholsj:
Making a new page which we can redirect various urls to (eg forum/apps/commenting etc) - we won't link to this page from anywhere else. 

Content for the page will come later. 

https://trello.com/c/HUBHaeoJ/227-page-for-retired-functionality


app/controllers/pages_controller.rb
```diff
@@ -16,6 +16,9 @@ def support
   def accessibility
   end
 
+  def site-changes
```
andrewgarner: The convention is to use underscores, so this should be `site_changes`.

app/views/pages/site-changes.html.erb
```diff
@@ -0,0 +1,14 @@
+<% content_for :page_title do %>
```
andrewgarner: The path to this file should be `app/views/pages/site_changes.html.erb`

config/routes.rb
```diff
@@ -21,6 +21,7 @@
     get 'support'
     get 'terms'
     get 'dashboard'
+    get 'site-changes'
```
andrewgarner: `get 'site-changes', to: :site_changes` should allow you to maintain the desired path but use underscoring for the controller method.

nicholsj:
@andrewgarner I've made those changes

config/routes.rb
```diff
@@ -21,7 +21,7 @@
     get 'support'
     get 'terms'
     get 'dashboard'
-    get 'site-changes'
+    get 'site_changes'
```
andrewgarner: If you change this to `get 'site-changes', to: :site_changes` you can have the URL "/site-changes" which I believe is preferable.

nicholsj:
Thanks @andrewgarner, change made

PR #431, opened 2018-02-23
Dashboard improvements

maxf:
We're going to have an external dashboard (grafana probably) that will display all sorts of things. It'll be nice to show our dashboard stats on it.


app/views/pages/dashboard.json.erb
```diff
@@ -0,0 +1,7 @@
+{
+"num_datasets": "<%= @datasets_count %>",
```
andrewgarner: We should render the integers as integers not strings. How about using https://github.com/rails/jbuilder?

PR #437, opened 2018-02-27
Boost ranking of datasets based on their organisation category

andrewgarner:
This adds additional should clauses to the Elasticsearch query to boost the score if the dataset belongs to a ministerial department, non-ministerial department or local council.

See https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-bool-query.html
Trello https://trello.com/c/rhB9ZPMa

### DIFF
--- a/app/services/search/query.rb
+++ b/app/services/search/query.rb
@@ -182,6 +182,18 @@ module Search
       query[:query][:bool][:must] << format_filter(format_param)       if format_param.present?
       query[:query][:bool][:must] << licence_filter(licence_param)     if licence_param.present?
 
+      # If we have a must clause then a dataset will match the bool query even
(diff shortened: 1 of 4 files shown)


app/services/search/query.rb
```diff
@@ -284,6 +296,22 @@ def self.licence_filter(licence)
       }
     end
 
+    def self.organisation_category_filter(organisation_category, boost: 2)
```
laurentqro: should we extract the magic number `2` into a constant?

andrewgarner: I'm not sure this would be considered a magic number. I've mirrored the Elasticsearch domain - https://www.elastic.co/guide/en/elasticsearch/reference/6.x/mapping-boost.html where "boost" is a thing.

PR #442, opened 2018-02-28
Fix layout for static pages

andrewgarner:
Sets correct heading styles and ensures the text in static pages is correctly constrained.

### DIFF
--- a/app/views/pages/about.html.erb
+++ b/app/views/pages/about.html.erb
@@ -1,5 +1,5 @@
 <% content_for :page_title do %>
-  About data.gov.uk -
+  About -
 <% end %>
 
 <main role="main" id="content">
--- a/app/views/pages/about.html.erb
+++ b/app/views/pages/about.html.erb
@@ -3,51 +3,66 @@
 <% end %>
 
 <main role="main" id="content">
-  <%= render partial: 'layouts/feedback_survey_banner' if flash[:beta_message] == 'hide'%>
+  <%= render partial: 'layouts/feedback_survey_banner' if flash[:beta_message] == 'hide' %>
 
   <div class="text">
 
-  <h1 class="heading-large">
-    About data.gov.uk
-  </h1>
+    <h1 class="heading-large">
+      About data.gov.uk
+    </h1>
 
-  <p>Since 2010 data.gov.uk has been helping people to find and use open government data, and supporting government publishers to maintain data. It’s the best place to freely publish and find government data.</p>
+    <p>Since 2010 data.gov.uk has been helping people to find and use open government data, and supporting government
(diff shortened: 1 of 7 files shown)


app/views/pages/about.html.erb
```diff
@@ -5,6 +5,8 @@
 <main role="main" id="content">
   <%= render partial: 'layouts/feedback_survey_banner' if flash[:beta_message] == 'hide'%>
 
+  <div class="text">
+
   <h1 class="heading-large">
```
maxf: can you fix the indentation inside that div?

app/views/pages/accessibility.html.erb
```diff
@@ -5,6 +5,8 @@
 <main role="main" id="content">
   <%= render partial: 'layouts/feedback_survey_banner' if flash[:beta_message] == 'hide'%>
 
+  <div class="text">
+
   <h1 class="heading-large">
```
maxf: likewise, etc.

app/views/pages/dashboard.html.erb
```diff
@@ -2,11 +2,11 @@
   What's available in Find
 <% end %>
 
-<h1 class="heading-large">
-  What's available in Find right now
-</h1>
-
 <main role="main" id="content">
+  <h1 class="heading-large">
+    What's available in Find right now
```
maxf: this doesn't look like an apostrophe.

app/views/pages/about.html.erb
andrewgarner: I purposefully didn't add the indentation so that the diff was readable.

There is unnecessary duplication between each page that I'd like to fix by refactoring out some shared layouts but I've saved that for a later date.

app/views/pages/dashboard.html.erb
andrewgarner: There is indifferent usage across all of our content. I didn't want to address that here when the focus is on heading sizes.

Incidentally referring to "Find" here is inconsistent with the reset of the site where we say "data.gov.uk" or "Find open data".

app/views/pages/about.html.erb
maxf: True it's harder to read, but it's a problem with git and it shouldn't dictate how we write our code. Proper indentation has saved us more than once. You can also reindent in a different commit, or just ignore whitespace changes in git diff

app/views/pages/dashboard.html.erb
maxf: Fair enough, I'd missed the fact it was already there.

app/views/pages/about.html.erb
andrewgarner: Point taken. Fixed with 84d72557f5052da8c8c2c742535a4f3ad8b7a263

PR #453, opened 2018-03-06
Update to header bar

hannako:
- trello card: https://trello.com/c/uDMK8vZY/194-add-datagovuk-to-black-header-bar-s
- add vertical separator between old and new service name in global header

# Before
screen shot 2018-03-06 at 13 37 41

# After 
screen shot 2018-03-06 at 13 36 34

### DIFF
--- a/app/assets/stylesheets/govuk-overrides.scss
+++ b/app/assets/stylesheets/govuk-overrides.scss
@@ -15,28 +15,34 @@
     img {
       display: none;
     }
+
     .vertical-separator {
       display: inline-block;
       margin-left: 4px;
       color: black;
       border-left: thin solid $grey-2;
-    }
-    .dgu-phase-tag--responsive {
-      @media (max-width: 641px) {
+      @media (max-width: 646px) {
         display: block;
-        width: 10%;
+        line-height: 0;
+        font-size: 0;
+      }
+    }
+
+    .dgu-phase-tag--responsive {
+      @media (max-width: 769px) {
+        display: block;
+        width: 12%;
+      }
+    }
+
+    .header-proposition {
+      #proposition-links {
+        a {
+          font-family: Helvetica;
+        }
       }
     }
   }
-
-  .header-proposition {
-    #proposition-links {
-      a {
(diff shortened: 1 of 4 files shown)


app/views/layouts/_vertical_separator.html.erb
```diff
@@ -0,0 +1,3 @@
+<svg id="vertical-separator" width="3" height="22">
+  <line x1="1.7" y1="0" x2="1.7" y2="22"></line>
+</svg>
```
maxf: We don#t normally use svg, because legacy browsers. You could use a left border on an inline-div?

maxf: We did it before, at some point, in one of the alphas. Blanca will remember.

andrewgarner:
We need to think how this will work responsively:

screen shot 2018-03-06 at 14 24 14

hannako:
@andrewgarner @maxf 

I can't replicate a screen size as small as you have in the screenshot. I've shrunk as far as possible in the chrome dev tools, and the only issue was the beta tag falling off the page. So now it's wrapping at screens smaller than 641.

andrewgarner:
@hannako I checked using the Responsive Design Mode in Safari and set it to iPhone SE. You can achieve similar in Chrome by setting it to "Responsive" then choosing "Mobile S".

hannako:
@maxf @andrewgarner 

For small screens, this is how it looks:

screen shot 2018-03-06 at 19 35 47

Happy with that?

laurentqro:
@BlancaTortajada ☝️ looks good?

BlancaTortajada:
@hannako @laurentqro 2 changes to be made please:
1. At the moment "Menu" dropdown is not floating on the right, just right aligned. Could you make it float so the Menu is aligned with the Beta tag?
2. The vertical spacing between the beta tag and "Find open data" should be the same as the spacing between "data.gov.uk" and "Find open data"

hannako:
@BlancaTortajada 

Max has requested we create a new card for your style notes and merge this as is for the time being.

PR #456, opened 2018-03-06
Place homepage under pages controller

laurentqro:
We place all our single non-resource pages under a `PagesController`. We apply this pattern in Publish as well.

To be consistent with this pattern, this PR places the home page under the `PagesController`.


config/locales/views/home/en.yml
```diff
@@ -1,5 +1,5 @@
 en:
-  home:
+  pages:
```
andrewgarner: By convention this file would also now be named `config/locales/views/pages/en.yml`

PR #468, opened 2018-03-14
Added new para: build services using registers

gtdata:
New content added


nicholsj:
Just adding a link to the Trello card https://trello.com/c/B6uCwKh7/309-a-way-of-signposting-registers-from-about-page

PR #471, opened 2018-03-14
Render custom message when user coming from legacy after dataset creation

laurentqro:
When a user creates a new dataset on legacy, they are redirected to the corresponding dataset page on Find.

However, because legacy -> Find sync is not instantaneous and we cannot modify legacy, we render a page informing the user their dataset will be available for view soon.

https://trello.com/c/A1LTiZQS/316-display-a-message-to-publishers-after-dataset-creation-m

### DIFF
--- a/app/views/datasets/_additional_info.html.erb
+++ b/app/views/datasets/_additional_info.html.erb
@@ -61,23 +61,6 @@
               <dd><%= i['metadata_language'] %></dd>
             <% end %>
           </dl>
-
-          <!--
-               Missing: (from https://data.gov.uk/dataset/lidar-for-scotland-phase-i)
-             - theme
-             - harvest URL
-             - harvest date
-             - Access constraints
-             - import source
-             - themes (secondary)
-             - temporal coverage
-             - schema/vocabulary
-             - code list
-             - service level
-             - coupled services
-             -->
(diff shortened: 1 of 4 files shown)


app/controllers/legacy/datasets_controller.rb
```diff
@@ -1,6 +1,20 @@
...
+  def available_soon
+    if request_originates_from_legacy?
+      render 'datasets/available_soon'
+    else
+      raise
+    end
+  end
+
+  def request_originates_from_legacy?
+    request.referrer.present? && URI.parse(request.referrer).host == ENV['LEGACY_HOST']
```
andrewgarner: Find and legacy will share the same host. I think we'll need to look at the path of the referrer.

app/views/datasets/_additional_info.html.erb
```diff
@@ -61,23 +61,6 @@
               <dd><%= i['metadata_language'] %></dd>
             <% end %>
           </dl>
-
-          <!--
```
andrewgarner: How is this related to this PR?

laurentqro: it isn't, just removing dead comments

PR #474, opened 2018-03-16
Replace text area with text field

laurentqro:
Text areas are resizeable, which is not an intended behaviour for name and email fields.

### DIFF
--- a/app/views/tickets/new.html.erb
+++ b/app/views/tickets/new.html.erb
@@ -55,10 +55,9 @@
               <span class="error-message">Enter a name</span>
             <% end %>
           </label>
-          <%= f.text_area(:name,
+          <%= f.text_field(:name,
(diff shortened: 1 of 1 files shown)


app/views/tickets/new.html.erb
```diff
@@ -54,7 +54,7 @@
               <span class="error-message">Enter a name</span>
             <% end %>
           </label>
-          <%= f.text_area(:name,
+          <%= f.text_field(:name,
                           class: input_box_class_for(@ticket, :name),
                           id: "example-name",
                           rows: 1 ) %>
```
andrewgarner: We should drop the rows attribute too.

laurentqro: yep, good spot

laurentqro: @andrewgarner removed the row attribute 👍

PR #495, opened 2018-03-23
328  jess/track previews

maxf:
This PR improves event handling for 3 types of events:
- clicking on an email link in the contact-publisher section
- clicking on a datafile link on a dataset page
- clicking on a datafile link on a preview page

All event data is added using data attributes. If a 'data-ga-event' attribute is found, then an onclick event handler is added that sends to ga the information contained in the other data-ga-* attributes and possibly href, as populated in the templates (publisher, link, etc).


app/views/datasets/_datafile_table.html.erb
```diff
@@ -38,7 +39,10 @@
           <% if datafile.html? %>
             <%= link_to t('.go_to_site'), datafile.url %>
           <% elsif datafile.csv? %>
-            <%= link_to t('.preview'), datafile_preview_path(@dataset.uuid, @dataset.name, datafile.uuid) %>
+            <%= link_to t('.preview'),
+                datafile_preview_path(@dataset.uuid, @dataset.name, datafile.uuid),
+                class: 'ga-preview'
```
hannako: doesn't this also need the `ga-event` class?

PR #498, opened 2018-03-27
Make page titles targets of jump-to-content links

maxf:
Previously links pointer to <main>, but it doesn't always start the actual content of a page, for instance when there's a feedback banner

https://trello.com/c/n3xWUfQx/334-skiplink-doesnt-skip-to-main-content-and-focus-remains-on-it

Reviewers please make sure that the dropdowns and autocomplete still work


hannako:
@maxf 

I think this PR has broken the layout on dataset show page and on the static pages (publishers and support)

PR #500, opened 2018-03-27
add <ul> tags to fix formatting of breadcrumbs

hannako:
JN made a comment on the absence of `<ul>` tags wrapping the `<li>` breadcrumbs that was introduced by https://github.com/alphagov/datagovuk_find/pull/478

It also introduced a minor change to the spacing:

Before PR 478

before

After PR 478

after

### DIFF
--- a/app/views/datasets/_breadcrumb.html.erb
+++ b/app/views/datasets/_breadcrumb.html.erb
@@ -4,19 +4,21 @@
     <div class="column-full">
       <div class="datagov_breadcrumb">
         <nav aria-label="Breadcrumb">
-          <li>
-            <%= link_to t('.home'), root_path %>
-          </li>
-          <li>
-            <% if @referer_query.nil? %>
-              <%= @dataset.organisation.title %>
-            <% else %>
-              <%= link_to t('.search'), "/search?#{@referrer}" %>
-            <% end %>
-          </li>
-          <li aria-current="page">
-            <%= shorten_title(@dataset.title) %>
-          </li>
+          <ol>
+            <li>
+              <%= link_to t('.home'), root_path %>
+            </li>
+            <li>
+              <% if @referer_query.nil? %>
+                <%= @dataset.organisation.title %>
+              <% else %>
+                <%= link_to t('.search'), "/search?#{@referrer}" %>
+              <% end %>
+            </li>
+            <li aria-current="page">
+              <%= shorten_title(@dataset.title) %>
+            </li>
+          </ol>
         </nav>
       </div>
     </div>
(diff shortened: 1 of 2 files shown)


app/views/datasets/_breadcrumb.html.erb
```diff
@@ -4,19 +4,21 @@
...
-            <% if @referer_query.nil? %>
-              <%= @dataset.organisation.title %>
-            <% else %>
-              <%= link_to t('.search'), "/search?#{@referrer}" %>
-            <% end %>
-          </li>
-          <li aria-current="page">
-            <%= shorten_title(@dataset.title) %>
-          </li>
+          <ul>
```
andrewgarner: I think this should be an ordered list (`ol`) as there is natural ordering of the breadcrumb elements.

This is the approach adopted in https://github.com/alphagov/govuk-frontend/blob/master/src/breadcrumbs/template.njk

hannako: Yeah I changed it from <ul> to <ol> after reading the W3 guidelines which suggested unless there's a number or similar it should be ol. But on the MDN docs it says different. https://developer.mozilla.org/en-US/docs/Web/HTML/Element/ol

Is it generally better to use the MDN docs?

hannako:
Accidentally closed!

app/views/datasets/_breadcrumb.html.erb
andrewgarner: > As a rule of thumb to determine which one to use, try changing the order of the list items; if the meaning is changed, the <ol> element should be used, else the <ul> is adequate.

app/views/tickets/new.html.erb
```diff
@@ -8,9 +8,11 @@
 
       <div class="datagov_breadcrumb">
         <nav aria-label="Breadcrumb">
-          <li>
-            <%= link_to 'Back', support_path %>
-          </li>
+          <ol>
+            <li>
+              <%= link_to 'Back', support_path %>
```
andrewgarner: This isn't directly related to your changes here, but perhaps this should be "Support" since we know that will always be the parent page. We're currently saying "Back" is a breadcrumb which isn't strictly true.

hannako: Sounds sensible. I'd have to check with Blanca. That was the [design](https://projects.invisionapp.com/d/main#/console/10081578/266102083/preview)

Might take a while. So suggest I make a new ticket.

andrewgarner: 👍

andrewgarner: If we stick with the current design we probably shouldn't give it an aria label of breadcrumb.

hannako: OK. I'll copy your comments on to the [trello card](https://trello.com/c/BuhLtrXU/241-fix-role-must-be-one-of-the-valid-aria-roles-rolebreadcrumb ) as it was a specific request from the accessibility audit. thanks

PR #501, opened 2018-03-27
add topics to the search results page title

hannako:
[trello card #331](https://trello.com/c/Cfv99DZM/331-search-results-page-title-from-category-search-should-be-more-descriptive)

Make the search results page title more descriptive when search by topic

### DIFF
--- a/app/views/search/search.html.erb
+++ b/app/views/search/search.html.erb
@@ -1,5 +1,9 @@
 <% content_for :page_title do %>
-   <%= t('.results_page_title', :query => @topic ) %>
+  <% if search_term.present? %>
+   <%= t('.results_page_title', :query => search_term ) %>
+  <% else %>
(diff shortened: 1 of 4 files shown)


app/views/search/search.html.erb
```diff
@@ -1,5 +1,5 @@
 <% content_for :page_title do %>
-   <%= t('.results_page_title', :query => @query ) %>
+   <%= t('.results_page_title', :query => @topic ) %>
```
andrewgarner: We should introduce some extra test coverage so we're confident we have the correct behaviour for different contexts.

If you do a normal search, for example, https://find-data-pr-501.herokuapp.com/search?q=Data the title now says 'Results for "" - data.gov.uk"' because the topic is empty.

hannako: 👍

PR #504, opened 2018-03-28
Updated content

gtdata:
Trello: https://trello.com/c/A1LTiZQS/316-display-a-message-to-publishers-after-dataset-creation-m


gtdata:
@andrewgarner not urgent, but the URL should align with the page title, is that a pain to do?

andrewgarner:
@gtdata the URL will be the legacy URL of the dataset, for example `https://test.data.gov.uk/dataset/my-dataset?utm_source=ckan#tab-name`

PR #520, opened 2018-04-06
Adds 'Preview on map' to WMS resources

rossjones:
Where a dataset has a WMS resource, and also has the appropriate INSPIRE
metadata, this PR will add a link to the legacy (data.gov.uk) map
preview for that WMS.

The map-based-preview requires the urlencoded `url` for the WMS
resource as well as the  `wfsurl` or any WFS resource in the same
dataset.  The bounding box (n, w, e, and s) is taken from the bbox_*
parameters of the INSPIRE metadata.

Not all WMS endpoints are available, so on visiting the map preview it's
possible that there will be:

* No layers to select
* A red 'x' over the WMS URL on the left hand side
* Errors when loading after selecting a layer
* No feature data for some layers

The PR works okay for the 30-something datasets I tried it on, apart
from those WMS servers that are borked. Luckily (or unluckily if JS is
disabled) the map preview uses JS to navigate back to the previous page
so we shouldn't need to modify the legacy map-preview.

### DIFF
--- a/app/views/datasets/_datafile_table.html.erb
+++ b/app/views/datasets/_datafile_table.html.erb
@@ -48,6 +48,8 @@
                   :'ga-publisher' => @dataset.organisation.name
                   }
             %>
+          <% elsif datafile.wms? && @dataset.inspire_dataset.present? %>
+            <%= link_to t('.map_preview'), map_preview_url(@dataset, datafile) %>
           <% else %>
             <span class="dgu-secondary-text"><%= t('.not_available') %></span>
           <% end %>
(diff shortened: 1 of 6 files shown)


app/views/datasets/_datafile_table.html.erb
```diff
@@ -48,6 +48,8 @@
                   :'ga-publisher' => @dataset.organisation.name
                   }
             %>
+          <% elsif datafile.wms? && @dataset.inspire_dataset.present? %>
+            <a href="<%= map_preview_url @dataset, datafile %>">Preview on map</a>
```
maxf: could you use t('.preview') ?

PR #524, opened 2018-04-09
DRY up feedback survey banner

laurentqro:
* Move duplicated view code to application layout
* Remove grey background for the search box


benthorner:
Blanca says she's happy to lose the gray backgrounds. Apparently these are going to be re-done in the new design system anyway.

PR #526, opened 2018-04-10
Use Yarn for accessible autocomplete styles

benthorner:


andrewgarner:
Rails 5.1 has Yarn support, we should follow that approach if we want to configure assets management.

benthorner:
Cloudfoundry only seems to have a buildpack for Nodejs/NPM. There's no actual difference between using NPM or Yarn in this case, as we're only using it as a package manager. So I think we may as well stay with what we've got, as it's still a net improvement.

package.json
```diff
@@ -0,0 +1,9 @@
+{
```
andrewgarner: Rails would generate the following:

```json
{
  "name": "datagovuk_find",
  "private": true,
  "dependencies": {
      "accessible-autocomplete": "^1.6.1"
  }
}
```

I'm not sure we need the `license`, `repository` and `version` properties.

.travis.yml
```diff
@@ -6,31 +6,36 @@ env:
...
-- elasticsearch
+  - postgresql
+  - elasticsearch
 addons:
   chrome: stable
 before_install:
-- gem update --system
+  - gem update --system
+install:
+  - bundle install --path vendor/bundle
```
andrewgarner: Did you try this without the explicit install steps? I __think__ should perform these steps when it sees a `Gemfile.lock` and `yarn.lock`

README.md
```diff
@@ -1,32 +1,39 @@
...
-## ElasticSearch
+Now run the following commands to complete your setup.
 
-If you want to run ElasticSearch on your development machine:
-* `brew install elasticsearch`
-* `elasticsearch`
+```
+bundle install
+rake db:setup
+yarn install
```
andrewgarner: We should add this step to https://github.com/alphagov/datagovuk_find/blob/master/bin/setup

```
# Install JavaScript dependencies if using Yarn
system('bin/yarn')
```

See https://github.com/rails/rails/blob/master/railties/lib/rails/generators/rails/app/templates/bin/setup.tt

andrewgarner:
We should try a manual deploy of this branch before merging to verify that yarn behaves correctly with the deployment pipeline.

package.json
benthorner: NPM needs all three, but turns out yarn only needs the license, so I'll remove the other two.

.travis.yml
benthorner: I did try without and the build failed. I think we both read the same doc, but the auto-yarn only works if the language is node.

benthorner:
Manual deploy is on my list of things to do :-)

README.md
benthorner: I prefer to go for the 'knowledge is power' approach and avoid setup scripts, as it makes everything seem a bit magic when you're learning. This one is quite a lot longer than 3 lines as well(!) - do we need it for something, or could people just what's in the README?

benthorner:
Test was a success @andrewgarner. Could you have a final check?

.travis.yml
```diff
@@ -6,32 +6,38 @@ env:
...
+  - curl -L https://codeclimate.com/downloads/test-reporter/test-reporter-latest-linux-amd64
+    > ./cc-test-reporter
+  - chmod +x ./cc-test-reporter
+  - "./cc-test-reporter before-build"
 script:
-- bundle exec brakeman
-- bundle exec rake db:create
-- bundle exec rake db:migrate
-- bundle exec rake spec
+<<<<<<< HEAD
```
andrewgarner: We don't want to include this.

How about we rebase this branch and rework the commits now that we've figured out an approach that works?

bin/setup
```diff
@@ -20,11 +20,6 @@ chdir APP_ROOT do
   system! 'gem install bundler --conservative'
   system('bundle check') || system!('bundle install')
 
-  puts "\n== Copying sample files =="
```
andrewgarner: How is this related to Yarn and accessible autocomplete?

.travis.yml
benthorner: Thanks for spotting - this is the n'th rebase and I must have missed it.

bin/setup
benthorner: This PR came from when I first started working on the project, so it partly turned into improving the README/setup so that it's easier. I've never run bin/setup on a rails project before and thought it was hiding a bit too much magic that should be in the README.

.travis.yml
benthorner: You're right that there is a lot on this branch, though. I'll try and get them into a separate PR.

benthorner:
I've extracted all of the README changes to a separate PR. This should be easier to review now :-)

PR #534, opened 2018-04-11
Override NTA with Helvetica

laurentqro:
We don't have the licence to use NTA, so we use helvetica instead.

* Override NTA with Helvetica
* Make sure message is no wider than 75 chars

### DIFF
--- a/app/views/layouts/_beta_message.html.erb
+++ b/app/views/layouts/_beta_message.html.erb
@@ -1,7 +1,7 @@
 <div class="dgu-beta__banner">
   <div class="dgu-beta__message">
     <p>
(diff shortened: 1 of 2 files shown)


app/views/layouts/_beta_message.html.erb
```diff
@@ -1,7 +1,7 @@
 <div class="dgu-beta__banner">
   <div class="dgu-beta__message">
     <p>
-      We’ve been improving data.gov.uk to help you find and use open government data.
+      We’ve been improving data.gov.uk to help you find and use open government data.<br />
```
andrewgarner: I think there's a `text` class we can add apply instead of the break. We use it on the static pages.

PR #536, opened 2018-04-12
Remove spurious "data.gov.uk" from search page titles

benthorner:


benthorner:
OK. Fixed - we're using a combination of locales and inline page titles :-)

PR #540, opened 2018-04-12
Header fix menu

benthorner:


andrewgarner:
The styling when you hover over the service name isn't quite right:

screen shot 2018-04-13 at 10 45 20

benthorner:
Fixed - can you take another look?

app/assets/stylesheets/govuk-overrides.scss
```diff
@@ -1,20 +1,16 @@
...
-
-body {
-  font-family: $helvetica-regular;
-}
+#global-header {
+  padding: 5px 0;
 
-// Homepage header overides__________________
+  .header-global {
+    @media (min-width: 769px) {
```
andrewgarner: How about we use the `$desktop-breakpoint` variable from https://github.com/alphagov/govuk_frontend_toolkit/blob/2011e77436fccd31f32a0a975a2bf4cc3b961495/stylesheets/_conditionals.scss#L28 ?

benthorner: That's better - that must be where they got that magic number from originally!

PR #385, opened 2018-02-08
Implement Topics on Find Beta

TobyRet:
- Add topics filter
- Display topic on a dataset
- Add topics filter links on home page

Corresponding PR for Publish Beta (this deals with importing and indexing topics) - https://github.com/alphagov/datagovuk_publish/pull/511

Trello card - https://trello.com/c/Ot65Qefh/80-implement-topics-themes-l (refer to this for mockups)

PR #541, opened 2018-04-13
Download button on Preview page: add GA event and hide button if link is broken

hannako:
On the preview page, the intended behaviour is to display a preview and a download button if the datafile URL points to a CSV. If the URL is broken we display an error message to the user and hide the button.

There was a bug where by the button was being displayed alongside the error message when the CSV link was broken.

## before:
screen shot 2018-04-13 at 07 44 42

## now:
screen shot 2018-04-13 at 07 44 06


app/views/previews/show.html.erb
```diff
@@ -10,22 +10,21 @@
...
-      <% end %>
-      <%= link_to t('.download_file'),
-      @datafile.url,
-        class: 'button',
-        :data => {
-          :'ga-event' => 'preview',
-          :'ga-format' => (@datafile.format.presence || 'n/a').upcase,
-          :'ga-publisher' => @dataset.organisation.name
-        }
+        <%= link_to t('.download_file'),
```
andrewgarner: We're nesting the download button in a `@datafile.csv?` condition which made me think, "if I previewed a non-csv file I wouldn't be able to download it" 🤔 

The view has upstream knowledge of how it is used, and is therefore coupled to the following condition:

https://github.com/alphagov/datagovuk_find/blob/562238f5b3d2dc72101ff2523f2df59e839c8f41/app/views/datasets/_datafile_table.html.erb#L40-L55

`Datafile` will attempt to generate a preview for any type of file:

https://github.com/alphagov/datagovuk_find/blob/562238f5b3d2dc72101ff2523f2df59e839c8f41/app/models/datafile.rb#L53-L55

But `Preview` only handles CSV files:

https://github.com/alphagov/datagovuk_find/blob/562238f5b3d2dc72101ff2523f2df59e839c8f41/app/models/preview.rb#L32

By using the `<% if @datafile.csv? && @datafile.preview.exists? %>` condition we're effectively leaking knowledge of the private behaviour of `Preview`.

This doesn't directly effect us now, but it makes it harder to add additional previews in the future.

app/views/previews/show.html.erb
```diff
@@ -10,22 +10,21 @@
...
+        @datafile.url,
+          class: 'button',
+          :data => {
+            :'ga-event' => 'preview',
```
andrewgarner: Should we be sending a 'preview' event for a download?

app/views/previews/show.html.erb
```diff
@@ -10,22 +10,21 @@
...
+            :'ga-format' => (@datafile.format.presence || 'n/a').upcase,
```
andrewgarner: This can only ever be "CSV".

app/views/previews/show.html.erb
rossjones: I think ALL of the preview stuff needs looking at when we need to support more than just CSV, but also we need to fix the `Preview` model because it is downloading the entire CSV into RAM if the remote server doesn't support HTTP-Range.  Although most CSV are less than 10k, some are very large.  I'll add a ticket for this.

andrewgarner: 👍

app/views/previews/show.html.erb
rossjones: @nicholsj is it possible to check GA and see whether this is doing the right thing?  Not sure what events we're expecting to store in GA.

nicholsj: Desired behaviour: 

Download : Category = datafile, action = download, label= <datafile url>

Preview :  Category = datafile-preview, action = click, label= <datafile url>
https://trello.com/c/WLA2YUze/328-add-ga-event-tracking-to-dataset-preview-links-s

What I'm seeing in GA:

Clicking the preview does send an event as desired. 
But clicking the download button from the preview page sends a preview event not a download one.

app/views/previews/show.html.erb
hannako: Thanks Andrew. Previews all need a refactor I agree. One option I discussed with laurent was if we run a link checker task periodically on the datafile URL's, and in Publish when the url is added, we can store  in the db `broken: true` or `false` then we wouldn't have to check the format or call `.preview` on each instance of `Datafile` either. Anyhow I'll make a ticket for it as it's out of scope of this PR.

andrewgarner: 👍

app/views/previews/show.html.erb
hannako: @nicholsj 
cc @rossjones 

There is a download event triggered by clicking download on the datafile table (on the dataset show page). How do you want to differentiate the download from the preview page? Are you happy with `Datafile-preview, action = download`

app/views/previews/show.html.erb
hannako: 👍

app/views/previews/show.html.erb
hannako: [trello ticket #384](https://trello.com/c/YIbDaNb8/384-linkchecker-for-datafile-urls)
[trello ticket #368](https://trello.com/c/pGP6cTkm/368-find-preview-assumes-everyone-support-http-range)

app/views/previews/show.html.erb
hannako: [Trello card #385](https://trello.com/c/ih1BdKOC/385-add-ga-datafile-download-event-on-preview-page)

PR #543, opened 2018-04-13
Replace arrow function (not supported in IE11)

laurentqro:
IE11 does not support arrow function, so on the dataset page the datafiles grouped by year under the show/hide button could not get unfolded.

This PR fixes this.

There is only one occurrence of the `=>` in the codebase. We are still getting an error in IE8, but the site remains functional.

https://trello.com/c/ZRUKFRpL/366-urgent-ie-bug-in-year-expando

paired with @emmabeynon

### DIFF
--- a/app/assets/javascripts/govuk.js
+++ b/app/assets/javascripts/govuk.js
@@ -125,7 +125,7 @@ FoldableText.prototype.toggle = function (event) {
 }
 
 FoldableText.prototype.init = function () {
-  $.each(this.els, (idx, el) => {
+  $.each(this.els, function(idx, el) {
     var $el = $(el)
     $el.css('max-height', '100000px')
     var originalHeight = $el.height()


rossjones:
Looks like this may have unintentionally broken the behaviour on normal browsers. https://find-data-pr-543.herokuapp.com/dataset/cd60f6d6-56bd-4511-90d0-31e012430f6c/wirral-expenditure-exceeding-500 doesn't look the same as https://data.gov.uk/dataset/cd60f6d6-56bd-4511-90d0-31e012430f6c/wirral-expenditure-exceeding-500 in FF.

laurentqro:
I suspect this has more to do with the fact that the review apps on Heroku are backed by a database with different data. In this occurrence, I think it's because the datafiles have no start date, which is what they get grouped by.

rossjones:
Okay, let's try it on staging.

PR #550, opened 2018-04-13
Render summaries as markdown

rossjones:
.. because they're markdown.

Currently all of the summaries are markdown, but ignore that and end up
squashed into a single paragraph/wall-of-text.

This PR renders the markdown to the page which means it will autolink,
and keep the appropriate spacing.

We can turn off certain features, so we can stop people using some
markdown features.

https://trello.com/c/HoH94aZv/363-markdown-in-dataset-description


rossjones:
@nicholsj This PR is just one so that we can see what happens when we render the summaries correctly - maybe we can discuss it some time next week?

app/helpers/datasets_helper.rb
```diff
@@ -29,6 +29,11 @@ def displayed_date(dataset)
     end
   end
 
+  def to_markdown(content)
+    markdown = Redcarpet::Markdown.new(Redcarpet::Render::HTML, autolink: true, tables: false)
```
andrewgarner: @rossjones did you look at using https://kramdown.gettalong.org/rdoc/ – it's the library used by https://github.com/alphagov/govspeak

rossjones: Ah no, I just used redcarpet through familiarity (I've used it before).
We probably should switch for consistency

andrewgarner: Another option maybe would be `Govspeak::Document.new(content).to_sanitized_html`

hannako: @rossjones yes no probs

@andrewgarner ta

PR #563, opened 2018-04-18
Hide summary HTML tag for IE

laurentqro:
No versions of Internet Explorer support the `<summary>` HTML tag (https://www.w3schools.com/tags/tag_summary.asp), so we hide it for them.

https://trello.com/c/vGOGy2B6/39-in-ie11-view-additional-metadata-link-does-nothing-as-metadata-is-already-visible

### Before (clicking on the link has no effect)

screen shot 2018-04-18 at 11 35 11

### After

screen shot 2018-04-18 at 11 34 49


benthorner:
Would this work instead...https://github.com/fnando/browser?

PR #565, opened 2018-04-18
Specify dataset's specific licence instead of blanket UK OGL

laurentqro:
Not all datasets fall under the UK OG licence, so this PR makes sure we specify the dataset's licence rather than applying UK OGL to all of them.

https://trello.com/c/WWUOqxyP/43-dcrights-metatag-is-shown-as-open-government-license-for-all-datasets


rossjones:
I think @andrewgarner is working on licence stuff in find at the moment, so would you mind too much if we left this PR for a while to look at afterwards?

nicholsj:
As mentioned, if showing the correct license in this metadata field is too fiddly (@andrewgarner has been battling with this for the front end display in Find) then we at least need to remove the ds:rights tag entirely in the short term

andrewgarner:
This should be fixed by https://github.com/alphagov/datagovuk_find/pull/568/

andrewgarner:
@laurentqro is their specific steer on whether we should use "dc:rights" or "dc:license"?

andrewgarner:
This is superseded by #568

PR #570, opened 2018-04-20
Change feedback/support links

nicholsj:
https://trello.com/c/Jql5wdEp/45-improve-signposting-for-feedback-vs-bug-reports-support

Directing feedback to the survey and problem reports to the support form.

### DIFF
--- a/app/views/pages/support.html.erb
+++ b/app/views/pages/support.html.erb
@@ -19,7 +19,7 @@
       </p>
 
       <p>
-        If you want to ask for a dataset, choose the data request option.
+        If you want to ask for a dataset, choose the data request option below.
       </p>
 
       <p>
--- a/app/views/pages/support.html.erb
+++ b/app/views/pages/support.html.erb
@@ -26,6 +26,11 @@
         For questions about a dataset, you can contact the publisher directly if they’ve provided contact details on the
         dataset page.
       </p>
+
+      <p>
+        If you have feedback about the site, use <%= link_to "this survey", "http://www.smartsurvey.co.uk/s/3SEXD/" %>.
+      </p>
+
       <p>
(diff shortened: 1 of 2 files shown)


app/views/pages/support.html.erb
```diff
@@ -26,6 +26,11 @@
         For questions about a dataset, you can contact the publisher directly if they’ve provided contact details on the
         dataset page.
       </p>
+
+      <p>
+        If you have feedback about the site, fill in <%= link_to "the survey", "http://www.smartsurvey.co.uk/s/3SEXD/" %>.
```
benthorner: This seemed a little punchy to me - I now feel less inclined to give my feedback ;-) Perhaps we could add some fluff?

app/views/pages/support.html.erb
```diff
@@ -26,6 +26,11 @@
         For questions about a dataset, you can contact the publisher directly if they’ve provided contact details on the
```
benthorner: The line above this talks about a 'data request option' - but I can't see one on the page. Is it somewhere else?

rossjones: Second option in form at https://data.gov.uk/support

benthorner: OK. It's a bit strange as I'm reading top-to-bottom and there's nothing in the text that talks about the form explicitly. Maybe we could add 'below' as a pointer?

nicholsj: Added

app/views/pages/support.html.erb
nicholsj: The content is from Gurinder - I'll ask her to look at it

nicholsj: In the meantime, I think it would be good to get these changes through so we can divert feedback to the survey. Content tweaks can come later

PR #578, opened 2018-04-25
Move elastic helpers into helper file

benthorner:


andrewgarner:
👍 

You could go further and isolate these methods in a module as documented here - https://relishapp.com/rspec/rspec-core/v/3-7/docs/helper-methods/define-helper-methods-in-a-module

benthorner:
I hadn't see that mechanism before. As we don't need to restrict their usage, I don't think we need it here though - but good to know for the times when we do.

PR #581, opened 2018-04-25
Use Rails 5-2-stable branch

laurentqro:
After upgrading Rails to 5.2, we saw `upstream sent too big header while reading response header from upstream` warnings in nginx logs, as well as `502 Bad gateway` errors.

The very likely cause: https://github.com/rails/rails/issues/32597.

A fix (https://github.com/rails/rails/pull/32602) has been backported into `5-2-stable` branch of Rails.

See change log (https://github.com/rails/rails/blob/5-2-stable/actionpack/CHANGELOG.md):

> Output only one Content-Security-Policy nonce header value per request.
> Fixes #32597.

This PR specifies the use of this branch.


benthorner:
Can you show that it's fixing the error in that article?

laurentqro:
@benthorner updated the description with proof of fix.

benthorner:
I feel like we should still demonstrate the problem is fixed. Can we do that with a simple curl?

PR #583, opened 2018-04-25
Change dataset 'Last updated' value

emmabeynon:

On the search results page each dataset displays a 'Last updated' date based on
its `last_updated_at` value. However after clicking through to a dataset page,
in cases where the dataset has datafiles, the metadata shows a 'Last updated'
date using the most recent datafile's `created_at` value, causing a mismatch.
This changes `Last updated` on dataset pages to display the most recent
datafile's `updated_at` value instead.

**Results page**

search_results_before

**Dataset page before**

dataset_page_before

**Dataset page after**

dataset_page_after

### DIFF
--- a/app/helpers/datasets_helper.rb
+++ b/app/helpers/datasets_helper.rb
@@ -23,7 +23,7 @@ module DatasetsHelper
     if dataset.datafiles.none?
       dataset.last_updated_at
     else
-      most_recent_datafile(dataset).created_at
+      most_recent_datafile(dataset).updated_at
     end
   end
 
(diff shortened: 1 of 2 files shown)


andrewgarner:
I think we also need to check that the right date is being show in the search results. The card says:

> For the DfT Naptan dataset, last updated shows as 26 March 2018. 
But on the actual dataset page
https://data.gov.uk/dataset/ff93ffc1-6656-47d8-9155-85ea0b8f2251/national-public-transport-access-nodes-naptan it says 12 June 2014 (which is the same as the last added file data - which is what we expect)

I think that means last updated should be 12th June 2014 in the search results too.

andrewgarner:
Ahh, I've just seen that more clearly explained in a comment on the card:

> On search results and dataset pages, show Last updated using the date of the most recently added datafile, or if there are none, the date the metadata was last changed.

PR #584, opened 2018-04-25
Govuk lint

benthorner:
https://trello.com/c/mK411OSf/96-add-govuk-lint-to-find


app/helpers/datasets_helper.rb
```diff
@@ -85,7 +85,7 @@ def to_json_ld(dataset)
       dataset_metadata[:keywords] = dataset.topic['title']
     end
     files = metadata_files(dataset)
-    if files.length > 0
+    if files.empty?
```
emmabeynon: As the logic is looking for files, this should be `unless`.

PR #564, opened 2018-04-18
Bump Rails from 5.1.6 to 5.2

laurentqro:
* upgrade Rails - http://weblog.rubyonrails.org/2018/4/9/Rails-5-2-0-final
* replace secure_headers gem (related: https://github.com/alphagov/datagovuk_find/pull/516)

PR #585, opened 2018-04-25
Revert to Rails 5.1.6 from 5.2

laurentqro:
Reverts https://github.com/alphagov/datagovuk_find/pull/564.

After upgrading Rails to 5.2, we saw `upstream sent too big header while reading response header from upstream` warnings in nginx logs, as well as `502 Bad gateway` errors.

The very likely cause: https://github.com/rails/rails/issues/32597.


andrewgarner:
☹️

config/application.rb
```diff
@@ -38,6 +38,8 @@ class Application < Rails::Application
 
     config.elasticsearch = config_for(:elasticsearch)
 
+    config.ssl_options = { hsts: { expires: 1.week } }
```
laurentqro: porting this over from `secure_headers` so that we keep only CSP-related logic in `secure_headers` and let Rails deal with what it can, as much as possible.

PR #587, opened 2018-04-26
Scope search filter values to those available in the query results

laurentqro:
This PR makes sure the search filters are scoped to the search query results.

https://trello.com/c/pZt678Ds/75-filters-should-show-only-include-applicable-options

https://www.elastic.co/guide/en/elasticsearch/guide/master/_scoping_aggregations.html

### DIFF
--- a/spec/features/search_page_spec.rb
+++ b/spec/features/search_page_spec.rb
@@ -308,4 +308,30 @@ feature 'Search page', elasticsearch: true do
   def assert_search_results_headings(*headings)
     expect(all('h2 a').map(&:text)).to contain_exactly(*headings)
   end
+
+  scenario 'search filters are scoped to search query results' do
+    foo_dataset = DatasetBuilder
+                    .new
+                    .with_title('Foo')
+                    .with_publisher('Org 1')
+                    .with_topic({ name: 'fun', title: 'Fun' })
+                    .with_datafiles([{ format: 'CSV' }])
+                    .build
+
+    bar_dataset = DatasetBuilder
+                    .new
+                    .with_title('Bar')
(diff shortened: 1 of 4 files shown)


andrewgarner:
@benthorner would you able to check if the review apps are working correctly?

benthorner:
@andrewgarner Automatic review apps are disabled because we're limited to 5, which was causing problems when we had too many concurrent PRs. You can still login to Heroku and create one from there though. James is investigating getting us a better account.

andrewgarner:
@rossjones I've got the acceptance test working for this. 👀 ❓

spec/features/search_page_spec.rb
```diff
@@ -308,4 +308,30 @@ def assert_data_set_length_is(count)
...
+                    .with_publisher('Org 2')
+                    .with_topic({ name: 'trivia', title: 'Trivia' })
+                    .with_datafiles([{ format: 'PDF' }])
+                    .build
+
+    index(foo_dataset, bar_dataset)
+
+    search_for('foo')
+
+    expect(page).to have_select('Publisher', options: ['', 'Org 1'])
```
rossjones: I think it's probably worthing also checking what ISN'T in the select, as the point is to make sure that the bar_dataset related objects aren't shown in the dropdown.

andrewgarner: That's what the `options` attribute of the matcher does – https://www.rubydoc.info/github/jnicklas/capybara/Capybara/Node/Matchers#has_select%3F-instance_method

rossjones: Ah okay, in that case LGTM

PR #588, opened 2018-04-26
Fix when unreleased label is used 

rossjones:
Currently we mark something as Not Released based on the absence of data files, but in reality it should be the absence of any files.

This PR adds a released? method that checks whether there are any docs or datafiles.  The Not Released text is shown when this returns false (there are no docs or datafiles).

The edit link will send the user to 'unpublished' edit on legacy ONLY when there are no links at all in a document.

https://trello.com/c/EH3XLYDr/116-change-logic-for-deciding-a-dataset-is-unavailable

https://trello.com/c/TgWGAY8l/117-make-edit-link-on-unavailable-datasets-going-to-unpublished-edit-form

### DIFF
--- a/app/helpers/datasets_helper.rb
+++ b/app/helpers/datasets_helper.rb
@@ -10,10 +10,10 @@ module DatasetsHelper
 
   def edit_dataset_url(dataset)
     url = URI::HTTPS.build(host: 'data.gov.uk')
-    url += if !dataset.released?
-             '/unpublished/edit-item/'
-           else
+    url += if dataset.released?
              '/dataset/edit/'
+           else
+             '/unpublished/edit-item/'
            end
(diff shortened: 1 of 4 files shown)


app/helpers/datasets_helper.rb
```diff
@@ -10,7 +10,7 @@ module DatasetsHelper
 
   def edit_dataset_url(dataset)
     url = URI::HTTPS.build(host: 'data.gov.uk')
-    url += if dataset.datafiles.none?
+    url += if !dataset.released?
```
andrewgarner: How about you flip the conditions to avoid the negation?

```rb
url += if dataset.released?
  '/dataset/edit/'
else
  '/unpublished/edit-item/'
end
```

PR #589, opened 2018-04-27
Never used the unpublished dataset editor

rossjones:
Regardless of whether we consider a dataset as 'unpublished' (in legacy
terms) always send the user to the normal dataset editor.  Some
workflows involve saving a dataset without any files intending to come
back to it, but they then can't because they get sent to the wrong
editor.


benthorner:
@rossjones @andrewgarner Is there a way we can test this on staging before it goes live?

PR #591, opened 2018-05-01
Fix linting

benthorner:


config/initializers/secure_headers.rb
```diff
@@ -1,14 +1,14 @@
...
+    default_src: ["'none'"],
+    connect_src: ["'self'", "www.google-analytics.com"],
+    font_src: ["'self'", "data:"],
+    img_src: ["'self'", "www.google-analytics.com"],
+    manifest_src: ["'self'"],
+    media_src: ["'self'"],
+    object_src: ["'self'"],
+    script_src: ["'unsafe-inline'", "'self'", "www.google-analytics.com"],
+    style_src: ["'unsafe-inline'", "'self'"]
   }
```
maxgds: This double/single quote stacking looks pretty ugly - could we just tell the linter not to apply that %w rule here instead of re-coding the config?

benthorner: I think we should avoid ignoring files in rubocop, as the config will get out of date as other files change. For me, it's less confusing seeing this than %w('...'), as I can clearly see it's intentional.

maxgds: Fair enough!

PR #595, opened 2018-05-02
Allow long contact email addresses to wrap

maxgds:
A fix for the behaviour documented here:
https://trello.com/c/MJQ4Mrrc/151-check-error-on-display-of-contact-details

Before:
screen shot 2018-05-02 at 13 12 19

After:
screen shot 2018-05-02 at 13 12 54

### DIFF
--- a/app/assets/stylesheets/components.scss
+++ b/app/assets/stylesheets/components.scss
@@ -410,6 +410,7 @@ span.showHide-open-all {
     padding-left: 0;
     span {
       display: block;
+      word-wrap: break-word;
     }
     h3 {
       padding-bottom: 8px;


app/assets/stylesheets/components.scss
```diff
@@ -410,6 +410,8 @@ span.showHide-open-all {
     padding-left: 0;
     span {
       display: block;
+      -ms-word-break: break-all;
```
andrewgarner: What are your thoughts on using something like https://github.com/ai/autoprefixer-rails?

maxgds: Looks interesting - not something I've come across before, but it seems less dirty than littering up the css directly. I'll look into it.

app/assets/stylesheets/components.scss
```diff
@@ -410,6 +410,8 @@ span.showHide-open-all {
...
+      word-break: break-all;
```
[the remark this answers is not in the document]
maxgds: Turns out, it does! thanks, I'll switch to that for this instance.

app/assets/stylesheets/components.scss
maxgds: @andrewgarner Have asked around and it's used elsewhere in gds, so will look at implementing that if needed in future. In the short term, though, I've refactored based on Vanita's comments.

andrewgarner: 👍

PR #599, opened 2018-05-03
Display public updated at on dataset show page

emmabeynon:

This adds `public_updated_at` to the Dataset model so that we can access it in Elasticsearch, and conditionally uses this value as the "Last updated" date on the Dataset show page when it's present.  After deploying https://github.com/alphagov/datagovuk_publish/pull/602 and rebuilding the Elasticsearch index then we'll remove the conditional logic in the `displayed_date` helper so that we only use `public_updated_at`.

### DIFF
--- a/spec/features/dataset_show_page_spec.rb
+++ b/spec/features/dataset_show_page_spec.rb
@@ -199,24 +199,35 @@ feature 'Dataset page', elasticsearch: true do
       expect(page).to have_content('Topic: Not added')
     end
 
-    context 'When a dataset has no datafiles' do
-      it 'Last Updated field displays last_updated_at' do
+    context 'When public_updated_at is present on a dataset' do
+      scenario 'Last Updated field displays public_updated_at' do
         dataset = DatasetBuilder.new.build
         index_and_visit(dataset)
         expect(page).to have_content("Last updated: #{dataset['last_updated_at']}")
       end
     end
 
-    context 'When a dataset has datafiles' do
-      it 'Last Updated field is the most recent datafiles updated_at date' do
-        dataset = DatasetBuilder.new
-          .with_datafiles(DATA_FILES_WITH_START_AND_ENDDATE)
-          .build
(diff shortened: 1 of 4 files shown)


spec/features/dataset_show_page_spec.rb
```diff
@@ -199,25 +199,10 @@
...
-        dataset = DatasetBuilder.new
-          .with_datafiles(DATA_FILES_WITH_START_AND_ENDDATE)
-          .build
-        index_and_visit(dataset)
-        datafiles = dataset[:datafiles]
-        last = datafiles.sort_by { |datafile| datafile[:updated_at] }.last
-        date = Time.parse(last["updated_at"]).strftime("%d %B %Y")
-        expect(page).to have_content("Last updated: #{date}")
-      end
+    scenario 'Last Updated field displays public_updated_at' do
```
rossjones: Is it worth testing the other paths through `displayed_date` as well?

emmabeynon: I'd added a couple of extra tests @rossjones 👍

spec/features/dataset_show_page_spec.rb
```diff
@@ -199,24 +199,45 @@
...
+        dataset = DatasetBuilder.new.build
+        dataset.delete(:public_updated_at)
+        index_and_visit(dataset)
+        expect(page).to have_content("Last updated: #{dataset['last_updated_at']}")
+      end
+    end
+
+    context 'When public_updated_at is not present on a dataset with datafiles' do
+      scenario 'Last Updated field displays most recent datafile updated_at' do
+        datafile = [
```
benthorner: Can you reuse the existing datafile constants from the dataset_spec_builder?

benthorner: Also, datafile is technically datafileS - maybe it would be shorter to just have it as a hash and then to `.with_datafiles([datafile])`.

PR #600, opened 2018-05-04
Remove ActiveRecord as we're not using it

benthorner:
https://trello.com/c/JZmeJrtx/159-create-a-e2e-integration-environment-on-govuk-heroku

Removing some of the database.yml config has created a bug. Rather than
fix this bug, it makes more sense to just remove ActiveRecord
completely.


andrewgarner:
I looked into this a while back and it breaks Kaminari.

benthorner:
Works for me in dev - I think you just need to make sure ActiveModel is still there.

benthorner:
Travis should be fine now. PR app is working too, although I need to fix the ES_HOST var.

andrewgarner:
Can you get Elasticsearch integration working so we can verify https://datagovuk-find-review-pr-600.herokuapp.com/search?q=Test ?

benthorner:
OK. It is working, although your particular query won't return any results as the Heroku apps all point to the master ES instance, which hasn't been filled with any data except one I published manually.

benthorner:
Definitely - will do.

PR #601, opened 2018-05-08
Display public updated at in search results

emmabeynon:

This updates the "recent" sort functionality to sort by `public_updated_at` instead of `last_updated_at`.  In cases where a dataset has datafiles, `public_updated_at` is the date a datafile was last updated, and in cases where a dataset has no datafiles, `public_updated_at` is the date the metadata was last updated.

Now that we have `public_updated_at` we no longer need the `displayed_date` helper to determine the correct date to display, so we remove it and updated references to it to use `public_updated_at` instead.

### DIFF
--- a/spec/features/dataset_show_page_spec.rb
+++ b/spec/features/dataset_show_page_spec.rb
@@ -199,36 +199,10 @@ feature 'Dataset page', elasticsearch: true do
       expect(page).to have_content('Topic: Not added')
     end
 
-    context 'When public_updated_at is present on a dataset' do
-      scenario 'Last Updated field displays public_updated_at' do
-        dataset = DatasetBuilder.new.build
-        index_and_visit(dataset)
-        expect(page).to have_content("Last updated: #{dataset['last_updated_at']}")
-      end
-    end
-
-    context 'When public_updated_at is not present on a dataset with no datafiles' do
-      scenario 'Last Updated field displays dataset last_updated_at' do
-        dataset = DatasetBuilder.new.build
-        dataset.delete(:public_updated_at)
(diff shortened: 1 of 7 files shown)


spec/features/dataset_show_page_spec.rb
```diff
@@ -199,36 +199,10 @@
...
-          .build
-        dataset.delete(:public_updated_at)
-
-        index_and_visit(dataset)
-        expect(page).to have_content("Last updated: #{datafile_updated_at}")
-      end
+    scenario 'Last Updated field displays public_updated_at' do
+      dataset = DatasetBuilder.new.build
+      index_and_visit(dataset)
+      expect(page).to have_content("Last updated: #{dataset['last_updated_at']}")
```
andrewgarner: Should this be `dataset['public_updated_at']`?

emmabeynon: Whoops, good spot!

spec/features/search_page_spec.rb
```diff
@@ -32,13 +32,12 @@
     old_dataset = DatasetBuilder.new
                     .with_title('Old Interesting Dataset')
                     .with_name('old-dataset')
-                    .last_updated_at('2014-07-24T14:47:25.975Z')
+                    .public_updated_at(Time.now - 1)
```
benthorner: Could you explain what the `- 1` is for?

emmabeynon: To make it an older date than the new dataset's public_update_at.

benthorner: Oops, I see - thanks :-)

PR #602, opened 2018-05-10
Protect against spans appearing without content

maxgds:
https://trello.com/c/9F1ZPIF7/109-dataset-contact-information-displays-incorrectly-on-mobile

This PR makes spans conditional depending on whether there is content for them to display.

### DIFF
--- a/app/views/datasets/_contact.html.erb
+++ b/app/views/datasets/_contact.html.erb
@@ -1,3 +1,10 @@
+<%
+  if contact_email_is_email?(dataset)
+    mail = mail_to(contact_email_for(dataset), nil, {'data-ga-event': 'contact', 'data-ga-publisher': @dataset.organisation.name})
+  else
+    mail = link_to(contact_email_for(dataset), contact_email_for(dataset), {'data-ga-event': 'contact', 'data-ga-publisher': @dataset.organisation.name})
+  end
+%>
 <section class="contact">
   <h2 class="heading-medium">
     <%= t('datasets.contact.contact') %>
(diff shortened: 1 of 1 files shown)


app/views/datasets/_contact.html.erb
```diff
@@ -10,14 +17,8 @@
...
       <p>
-        <span><%= contact_name_for(dataset) %></span>
-        <span>
-          <% if contact_email_is_email?(dataset) %>
-            <%= mail_to(contact_email_for(dataset), nil, {'data-ga-event': 'contact', 'data-ga-publisher': @dataset.organisation.name}) %>
-          <% else %>
-            <%= link_to(contact_email_for(dataset), contact_email_for(dataset), {'data-ga-event': 'contact', 'data-ga-publisher': @dataset.organisation.name}) %>
-          <% end %>
-        </span>
+        <%= content_tag(:span, contact_name_for(dataset)) if contact_name_for(dataset) != "" %>
```
andrewgarner: Ruby has a `empty?` [method](https://ruby-doc.org/core-2.1.0/String.html#method-i-empty-3F) we could use here though it's more common place to use the `blank?` [method](http://api.rubyonrails.org/classes/Object.html#method-i-blank-3F) or the inverse `present?` method provided by Rails.

Using one of these makes the code a little more expressive:

```erb
<%= content_tag(:span, contact_name_for(dataset)) if contact_name_for(dataset).present? %>
```

app/views/datasets/_contact.html.erb
```diff
@@ -29,14 +30,13 @@
...
-        <span><%= mail_to(foi_email_for(dataset), nil, {class: 'ga-contact'}) %></span>
-        <span>
-          <%= link_to(foi_web_address_for(dataset),
-              foi_web_address_for(dataset),
-              {class: 'ga-contact'})
-          %>
-        </span>
+        <%= content_tag(:span, foi_name_for(dataset)) if foi_name_for(dataset) != "" %>
+        <%= content_tag(:span, mail_to(foi_email_for(dataset), nil, {class: 'ga-contact'})) if foi_email_for(dataset) != "" %>
+        <%=
```
andrewgarner: You can pass a block to the `content_tag` method which would be a good thing to do here:

```erb
<% if foi_web_address_for(dataset).present? %>
  <%= content_tag(:span) do %>
    <%= link_to(foi_web_address_for(dataset), foi_web_address_for(dataset), { class: 'ga-contact'} ) %>
  <% end %>
<% end %>
```

maxgds:
Adjusted as recommended, thanks. Please could you take another look @andrewgarner ?

PR #606, opened 2018-05-16
Refactor zendesk implementation with tests

benthorner:
https://trello.com/c/i5iX0OiN/224-zendesk-integration-is-broken

The previous implementation used a set of client classes (models?!),
assigned to a global constant, which introduced an unnecessary amount
of indirection. Instead, we now have a Zendesk module that can return a
real or dummy client as appropriate.

Note that Zendesk will now be enabled for all non-dev/test environments,
as we should allow for manual testing in these environments. This will require populating the appropriate environment variables with the production secrets.


spec/spec_helper.rb
```diff
@@ -18,21 +19,6 @@
     delete_index
   end
 
-  config.expect_with :rspec do |expectations|
-    expectations.include_chain_clauses_in_custom_matcher_descriptions = true
-  end
-
-  config.mock_with :rspec do |mocks|
```
andrewgarner: How is this related?

spec/spec_helper.rb
```diff
@@ -18,21 +19,6 @@
...
-    mocks.verify_partial_doubles = true
-  end
-
-  config.shared_context_metadata_behavior = :apply_to_host_groups
```
andrewgarner: Again, how is this related?

spec/spec_helper.rb
```diff
@@ -18,21 +19,6 @@
...
-
-  if config.files_to_run.one?
```
andrewgarner: Again, how is this related?

spec/spec_helper.rb
```diff
@@ -18,21 +19,6 @@
...
-    config.default_formatter = "doc"
-  end
-
-  config.profile_examples = 1
```
andrewgarner: Again, how is this related?

benthorner: Indeed, it's not. I'm trying to feel where the 'cleaning up a file can go in the same PR' line is.

PR #608, opened 2018-05-17
Update Rails and RSpec configs and generated files

andrewgarner:
This is proposed as an alternative to #607.

It brings the configs and files generated by `$ rails app:update` and `$ rails generate rspec:install` in line with the versions of Rails and RSpec we depend on.

This means that:
- we adopt the community driven conventions and opinions provided by the frameworks
- we minimise the diff when we repeat the process when upgrading to future versions of the frameworks


bin/setup
```diff
@@ -3,26 +3,32 @@ require 'pathname'
...
   puts '== Installing dependencies =='
   system! 'gem install bundler --conservative'
   system('bundle check') || system!('bundle install')
-  system! 'bin/yarn'
+
+  puts '== Installing JavaScript dependencies =='
+  system('bin/yarn')
 
   puts "\n== Setting environment =="
   unless File.exist?('.env')
```
benthorner: This should be unnecessary now.

benthorner:
Nooooo...pointless comments :-( But yes, it's a logical approach.

bin/setup
andrewgarner: Indeed, good spot!

benthorner:
In fact, we can remove .env, I think?

andrewgarner:
`.env` isn't tracked by git so removing the step from `bin/setup` should be good!

benthorner:
But also the gem :-)

benthorner:
dotenv-rails

PR #612, opened 2018-05-18
Replace DatasetBuilder with FactoryBot

benthorner:
https://trello.com/c/9IACvTIA/238-replace-custom-datasetbuilder-factor-with-factorybot

FactoryBot is the standard solution for creating model objects or testing. Using FactoryBot significantly reduces the size of the tests!

Some of the tests now make more use of the defaults specified in the factory. This helps keep the test size down and encourages making the defaults realistic.


spec/features/search_page_spec.rb
```diff
@@ -4,113 +4,71 @@
...
-                .build
-
+    dataset = build :dataset, title: 'a title'
     index(dataset)
-    search_for(query)
+    search_for('title')
 
     expect(page).to have_css('h1', text: 'Search results')
-    expect(page).to have_css('a', text: dataset_title)
+    expect(page).to have_css('a', text: 'title')
```
andrewgarner: This assertion is now very generic – the page should have an anchor containing the text "title". There is a risk this could pass for the wrong reason.

Are we asserting it should have a link to the search term — which is now "title" — or is it a sub match on the title of the dataset — which is now "a title" — we expect to be in the results?

spec/features/search_page_spec.rb
```diff
@@ -4,113 +4,71 @@
...
 
+    index(dataset1, dataset2)
     visit('/search')
-
     assert_data_set_length_is(2)
 
     execute_script "window.location = '#publisher'"
-
-    find_field('publisher').send_keys('Tob', :down, :enter)
+    find_field('publisher').send_keys(dataset1.organisation.title, :down, :enter)
```
andrewgarner: This test is simulating "Tob" being typed, and the autocomplete behaviour matching the full name of the publisher "Toby Corp". We've not lost that intent.

spec/features/search_page_spec.rb
benthorner: Fair enough - being concise was my aim.

spec/features/search_page_spec.rb
benthorner: Good point - done.

spec/features/search_page_spec.rb
```diff
@@ -4,63 +4,40 @@
...
-                    .with_name('old-dataset')
-                    .public_updated_at(Time.now - 1)
-                    .build
-
-    new_dataset = DatasetBuilder.new
-                    .with_title('Recent Interesting Dataset')
-                    .with_name('new-dataset')
-                    .build
-
+    old_dataset = build :dataset, title: 'Old Interesting Dataset', public_updated_at: 1.hour.ago
```
benthorner: The previous difference of 1 second caused the tests to flash.

spec/features/search_page_spec.rb
```diff
@@ -4,63 +4,40 @@
...
+    new_dataset = build :dataset, title: 'Recent Interesting Dataset'
     index(old_dataset, new_dataset)
 
     search_for('Old Interesting Dataset')
-
     expect(page).to have_css('option[selected]', text: 'Best match')
```
benthorner: The large number of new lines here was making the test hard to follow - removed a few to group the together a bit more.

spec/features/search_page_spec.rb
```diff
@@ -69,48 +46,29 @@
...
-                      .with_publisher('Tonka Trucks')
-                      .build
-
-    second_dataset = DatasetBuilder.new
-                      .with_title('Data About Toby')
-                      .with_publisher('Toby Corp')
-                      .build
-
-    index(first_dataset, second_dataset)
+    dataset1 = build :dataset, title: 'Wanted'
```
benthorner: Using factory default texts should make it easier to concentrate on the purpose of the test and less on the amusing content.

spec/features/search_page_spec.rb
```diff
@@ -121,165 +79,88 @@
...
+    dataset2 = build :dataset, :unrelated
 
+    index(dataset1, dataset2)
     visit('/search')
-
     assert_data_set_length_is(2)
 
     execute_script "window.location = '#format'"
-    find_field('format').send_keys('FO', :down, :enter)
+    find_field('format').send_keys('CSV', :down, :enter)
```
benthorner: Using factory defaults makes the test more realistic here.

spec/features/search_page_spec.rb
```diff
@@ -121,165 +79,88 @@
...
-                                             )
-                                             .build
-
-    index(
-      ministerial_dataset,
-      non_ministerial_dataset,
-      local_council_dataset,
-      non_departmental_public_body_dataset
-    )
+    dataset1 = build :dataset, organisation: build(:organisation, :raw)
```
benthorner: Note: this test is actually passing by chance. The organisation boost functionality is broken.

spec/features/search_page_spec.rb
```diff
@@ -121,165 +79,88 @@
...
-                                     .or eq('Forestry Commission')
-
-    expect(publishers[1]).to eq('Department for Environment Food & Rural Affairs')
-                               .or eq('English Heritage')
-                                     .or eq('Forestry Commission')
-
-    expect(publishers[2]).to eq('Department for Environment Food & Rural Affairs')
-                               .or eq('English Heritage')
-                                     .or eq('Forestry Commission')
+    publishers[0..2].each do |publisher|
```
benthorner: This is a no-brainer instead of changing something 3 times.

spec/requests/error_handling_spec.rb
```diff
@@ -15,20 +15,10 @@
   end
 
   it "handles a Datafile that can't be found by UUID" do
-    dataset_name = 'my-dataset'
-    dataset_uuid = SecureRandom.uuid
-    datafile_uuid = SecureRandom.uuid
```
benthorner: We could leave this variable in, but it wouldn't improve the diff.

benthorner:
The unnecessary changes should now be reverted and the diff a bit shorter. I've written a comment to justify each additional break from the original tests - they all made the change easier.

spec/features/search_page_spec.rb
benthorner: If you change the order the datasets are indexed in, the test fails.

spec/features/dataset_show_page_spec.rb
```diff
@@ -101,385 +61,229 @@
...
       expect(page).to have_content('Topic: Not added')
     end
 
     scenario 'Last Updated field displays public_updated_at' do
-      dataset = DatasetBuilder.new.build
+      dataset = build :dataset
+      time = Time.parse(dataset.public_updated_at)
       index_and_visit(dataset)
-      expect(page).to have_content("Last updated: #{dataset['public_updated_at']}")
+      expect(page).to have_content("Last updated: #{time.strftime('%d %B %Y')}")
```
benthorner: The builder was previously injecting an unrealistic (non-iso8601) string.

spec/features/dataset_show_page_spec.rb
```diff
@@ -101,385 +61,229 @@
...
-        .build
-
-      index_and_visit(dataset)
-
-      expect(page).to have_css('h2', text: 'Data links')
-    end
   end
 
   feature 'Side bar content' do
+    let(:dataset1) { build :dataset, title: '1 Data Set' }
```
benthorner: These datasets are common to all of the following scenarios, so it makes sense to extract them here instead of typing them over again. None of these scenarios require the dataset to have a datafile.

spec/features/dataset_show_page_spec.rb
```diff
@@ -122,24 +99,19 @@
...
-                  .with_licence_custom('For feature specs only.')
-                  .build
+      dataset = build :dataset, :with_custom_licence,
+                                licence_custom: 'For feature specs only.'
 
       index_and_visit(dataset)
 
       expect(page)
-        .to have_css('meta[name="dc:rights"][content="Feature Spec Open License 2.1"]',
+        .to have_css('meta[name="dc:rights"][content="Example Open License 1.1"]',
```
benthorner: The intention here is to re-use the factory trait for explicit/custom licences.

PR #614, opened 2018-05-18
Released strikes again

benthorner:
https://trello.com/c/SkQ9KxWf/139-show-availability-of-datasets-in-search-results

### DIFF
--- a/app/views/search/search.html.erb
+++ b/app/views/search/search.html.erb
@@ -50,7 +50,7 @@
                 <%= link_to dataset.title, dataset_path(dataset.uuid, dataset.name) %>
               </h2>
               <dl class="dgu-metadata__box">
-                <% if !dataset.released? %>
+                <% unless dataset.released? %>
                   <dt><%= t('.meta_data_box.availability') %>:</dt>
                   <dd class="availability">
                   <span class="dgu-highlight">Not released</span>
--- a/app/views/search/search.html.erb
+++ b/app/views/search/search.html.erb
@@ -50,6 +50,12 @@
                 <%= link_to dataset.title, dataset_path(dataset.uuid, dataset.name) %>
               </h2>
               <dl class="dgu-metadata__box">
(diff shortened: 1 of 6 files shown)


app/views/search/search.html.erb
```diff
@@ -50,6 +50,12 @@
                 <%= link_to dataset.title, dataset_path(dataset.uuid, dataset.name) %>
               </h2>
               <dl class="dgu-metadata__box">
+                <% if !dataset.released? %>
```
andrewgarner: I'm not sure this predicate method exists? Maybe we could change the implementation in `app/models/dataset.rb` so the revert commit applied cleanly and we don't have to change `app/views/datasets/_meta_data.html.erb`?

```rb
def released?
  released.present?
end
```

spec/features/dataset_show_page_spec.rb
```diff
@@ -518,38 +518,16 @@ def create_datafiles(count)
...
-    end
-
-    scenario 'Not released is shown when there are no docs and no files' do
-      dataset = DatasetBuilder
-                  .new
-                  .with_legacy_name('abc123')
-                  .build
-
+    scenario 'released is set to true' do
+      dataset = DatasetBuilder.new.build
```
andrewgarner: It might be safer to explicitly set `.with_released(true)` here so we're not relying on the default behaviour of `DatasetBuilder`.

spec/features/search_page_spec.rb
```diff
@@ -27,6 +27,16 @@
     expect(page).to have_css('a', text: dataset_title)
   end
 
+  scenario 'Search results indicate dataset availability' do
+    index(DatasetBuilder.new.build)
```
andrewgarner: It might be safer to explicitly set `.with_released(true)` here so we're not relying on the default behaviour of `DatasetBuilder`.

PR #641, opened 2018-06-22
Add option to support for to become a publisher

benthorner:
https://trello.com/c/bGmVDyz0/409-replace-registration-process

The breadcrumb now correctly shows the type of support being requested,
which incurred a refactoring to get this data from the ticket model,
rather than from an instance variable.

### DIFF
--- a/app/controllers/tickets_controller.rb
+++ b/app/controllers/tickets_controller.rb
@@ -1,7 +1,6 @@
 class TicketsController < ApplicationController
   def new
-    @support_queue = params['support']
-    @ticket = Ticket.new
+    @ticket = Ticket.new(support: params['support'])
   end
 
   def create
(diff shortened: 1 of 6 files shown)


gtdata:
Please add I want to create an account instead of I want to publish for an organisation (as that is what the user needs us to do to be able to publish data) 

Please update the dynamic page for this option too.
So Publish for an organisation becomes Create an account

Thanks,

benthorner:
@gtdata So the reason for saying 'publish for an organisation' is that we want to make sure the user mentions the organisation in their submission, without having to write a bespoke form with a field for it. Do you think they will put it in if we don't mention the word?

gtdata:
Ah okay, yes, stick to the original version if we aren't able to edit the existing fields.
Cheers

PR #642, opened 2018-06-22
Fix map preview CSP

andrewgarner:


app/controllers/map_previews_controller.rb
```diff
@@ -40,10 +43,12 @@ def correct_url(url)
     # http://lasigpublic.nerc-lancaster.ac.uk/ArcGIS/services/Biodiversity/GMFarmEvaluation/MapServer/WMSServer?request=GetCapabilities&service=WMS
```
benthorner: Can we remove the comment?

andrewgarner: It's an example that @rossjones noted. I'd prefer to keep it since this aspect of the application is somewhat opaque.

benthorner: Could we move it to the ADR when we do that?

andrewgarner: Perhaps yes.

PR #645, opened 2018-06-25
Allow map preview to load content images from any URL

andrewgarner:
Previously we tried to set a strong CSP so that map preview images could only be loaded from OS and the WMS host. This is proving to be too strict as we have WMS sources that link to other URIs for their map data. For example https://data.gov.uk/data/map-preview?e=1.8&n=55.8&s=50.0&url=http%3A%2F%2Fenvironment.data.gov.uk%2Fds%2Fwms%3FSERVICE%3DWMS%26INTERFACE%3DENVIRONMENT--6f59c1ce-cc11-43aa-b11d-e1c3ab43a192%26request%3DGetCapabilities&w=-5.7

This allows images to be loaded from any URI so that we can support map previews.

### DIFF
--- a/app/controllers/map_previews_controller.rb
+++ b/app/controllers/map_previews_controller.rb
@@ -4,16 +4,14 @@ require 'open-uri'
 
 class MapPreviewsController < ApplicationController
   def show
-    wms_uri = URI(url_param)
-    wms_img_src = "#{wms_uri.scheme}://#{wms_uri.host}"
-
     # rubocop:disable Lint/PercentStringArray
     append_content_security_policy_directives(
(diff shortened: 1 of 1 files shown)


benthorner:
Can you confirm all these URLs are coming from a trusted source?

nicholsj:
I think we have to override the CSP for map preview. We can't confirm that all URLs are trusted, but we have to trust that if it's in the WMS it's ok (not ideal but pragmatic - there was no CSP on legacy at all). 

The alternative (that map preview breaks) isn't an option - and we don't have resource to explore this further

benthorner:
That's OK - sounds like WMS is a trusted source :-)

PR #646, opened 2018-06-28
Revert "Remove scripts directory"

andrewgarner:
Reverts alphagov/datagovuk_find#635

These are required by the `deploy` step of the Travis config. A wrapper script is the documented approach – https://docs.travis-ci.com/user/deployment/script/

### DIFF
--- a/scripts/deploy-production.sh
+++ b/scripts/deploy-production.sh
@@ -1,0 +1,5 @@
+#!/bin/bash
+
+set -e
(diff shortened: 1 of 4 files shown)


benthorner:
Oops, I seem to have a habit of forgetting Travis. I'll look at a separate PR to get rid of the two redundant scripts (deploy.sh and set-env.sh).

PR #648, opened 2018-07-12
Do not bind GDS prometheus to the PaaS app

idavidmcdonald:
This application no longer requires metric collection by the
GDS prometheus, as implemented by
https://github.com/alphagov/datagovuk_find/pull/633

This commit removes the PaaS service bind so that Prometheus
will not try and collect metrics as intended.


idavidmcdonald:
Duplicate

PR #657, opened 2018-11-23
Redirects all unknown routes to CKAN

andreagrandi:
This is my attempt to redirect all unknown routes to CKAN url.

p.s: I would prefer to have that URL read from a ENV variable, but I've no idea how to do that

### DIFF
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -78,4 +78,6 @@ Rails.application.configure do
   if ENV["RAILS_LOG_TO_STDOUT"].present?
     config.logger = ActiveSupport::Logger.new(STDOUT)
   end
+
+  config.ckan_redirection_url = ENV['CKAN_REDIRECTION_URL']
 end
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -46,4 +46,9 @@ Rails.application.routes.draw do
   get 'dataset/:dataset_uuid/:name/datafile/:datafile_uuid/preview', to: 'previews#show', as: 'datafile_preview'
 
   get 'acknowledge', to: 'messages#acknowledge'
+
+  # Route everything else to CKAN
+  if ENV["CKAN_REDIRECTION_URL"].present?
+    match "*path", to: "pages#redirect", via: :all
+  end
 end
--- a/production-manifest.yml
+++ b/production-manifest.yml
@@ -10,6 +10,7 @@ applications:
   env:
     RAILS_ENV: production
     RACK_ENV: production
+    CKAN_REDIRECTION_URL: https://ckan.publishing.service.gov.uk
   services:
   - find-production-secrets
   - logit-ssl-drain
--- a/staging-manifest.yml
+++ b/staging-manifest.yml
@@ -10,6 +10,7 @@ applications:
   env:
     RAILS_ENV: staging
     RACK_ENV: staging
+    CKAN_REDIRECTION_URL: https://ckan.staging.publishing.service.gov.uk
   services:
   - find-staging-secrets
   - elasticsearch-beta-staging
--- a/app/controllers/pages_controller.rb
+++ b/app/controllers/pages_controller.rb
@@ -8,6 +8,10 @@ class PagesController < ApplicationController
     @datafiles_count_by_format = Dataset.datafiles['formats']['buckets']
   end
 
+  def redirect
+    redirect_to("#{ENV['CKAN_REDIRECTION_URL']}/#{params[:path]}")
+  end
+
 private
 
   def get_datasets_count
--- a/config/environments/development.rb
+++ b/config/environments/development.rb
@@ -50,4 +50,6 @@ Rails.application.configure do
   config.file_watcher = ActiveSupport::EventedFileUpdateChecker
 
   config.zendesk = nil
+
+  config.ckan_redirection_url = ENV['CKAN_REDIRECTION_URL']
 end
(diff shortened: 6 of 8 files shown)


config/routes.rb
```diff
@@ -46,4 +46,7 @@
   get 'dataset/:dataset_uuid/:name/datafile/:datafile_uuid/preview', to: 'previews#show', as: 'datafile_preview'
 
   get 'acknowledge', to: 'messages#acknowledge'
+
+  # Route everything else to CKAN
+  get '*path' => redirect('http://ckan.publishing.service.gov.uk/')
```
brucebolt: This app also runs in staging, so is it possible to route requests on staging to ckan.staging.publishing.service.gov.uk?  There are environment configurations in this app, so it may be possible to define the redirect URL in there.

Also, I think we should route to https by default.

config/environments/production.rb
```diff
@@ -78,4 +78,6 @@
   if ENV["RAILS_LOG_TO_STDOUT"].present?
     config.logger = ActiveSupport::Logger.new(STDOUT)
   end
+
+  config.ckan_redirection_url = ENV['CKAN_REDIRECTION_URL'].present?
```
brucebolt: This will only create the config variable in production.  It also needs to be specified in `config/environments/staging.rb`.

Additionally, `.present?` returns true or false, rather than the actual value.

config/routes.rb
```diff
@@ -48,5 +48,5 @@
   get 'acknowledge', to: 'messages#acknowledge'
 
   # Route everything else to CKAN
-  get '*path' => redirect('http://ckan.publishing.service.gov.uk/')
+  get '*path' => redirect(Rails.configuration.ckan_redirection_url)
```
brucebolt: Should this be wrapped in a `if` block?  I'm wondering what users would get redirected to if the config value isn't present (e.g. locally in development)?

config/environments/production.rb
brucebolt: Might it be better to just define the URL here rather than needing to use an environment variable?

andreagrandi: @brucebolt I'm removing the `.present?`, but I'm not sure about putting it in `staging.rb` because that file includes production: `require_relative './production'`
doesn't this mean that it will inherit all the values defined in production? There is a similar thing we do for Django config so I assumed this was the equivalent, but I may be wrong

About the env variable: I usually use them so whenever you need to fix a value there is no need to change the source code, you will just change the value in the environment. But if it's not something you normally do, no problem, I can use the fix value

config/routes.rb
andreagrandi: Yes, it makes sense, I'm going to fix this

andreagrandi: Done

brucebolt:
Looks good.  I agree with your point about the environment variable, it makes testing much easier (as I've just done on my local machine).

However just a few things:
* Could this also be applied to the `development` environment, so it works locally?
* Are there any tests we could add?
* Can you squash and rebase the commits so the history is cleaner?

config/environments/test.rb
```diff
@@ -39,4 +39,6 @@
 
   # Raises error for missing translations
   config.action_view.raise_on_missing_translations = true
+
```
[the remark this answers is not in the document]
andreagrandi: how do we set `ENV['CKAN_REDIRECTION_URL'] ||= 'http://testdomain'` for running unit tests then?

staging-manifest.yml
```diff
@@ -10,6 +10,7 @@ applications:
   env:
     RAILS_ENV: staging
     RACK_ENV: staging
+    CKAN_REDIRECTION_URL: https://ckan.staging.publishing.service.gov.uk/
```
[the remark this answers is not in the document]
andreagrandi: I'm not sure... I think @brucebolt mentioned it a few days ago, but maybe I can't remember it exactly?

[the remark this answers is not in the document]
brucebolt: I believe integration for Find is a Heroku app, therefore it doesn't have a CloudFoundry manifest.  The environment config for integration is specified in https://github.com/alphagov/datagovuk_find/blob/master/config/environments/integration.rb.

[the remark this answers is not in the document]
andreagrandi: Ok, I can remove it then.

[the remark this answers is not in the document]
brucebolt: No, keep it here for staging, but add additional config for the integration environment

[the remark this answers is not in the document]
andreagrandi: both staging and integration import production which has `config.ckan_redirection_url = ENV['CKAN_REDIRECTION_URL']` I thought this was fine for all of them?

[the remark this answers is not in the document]
brucebolt: But the environment variable must be set somewhere

config/environments/test.rb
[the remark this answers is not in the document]
elliotcm: It can go back into `rails_helper.rb`

staging-manifest.yml
brucebolt: I don't think we need the trailing slash, otherwise the user will be redirected to `https://ckan.staging.publishing.service.gov.uk///somewhere` (the slash is already defined on line 12 of `app/controllers/pages_controller.rb`).

production-manifest.yml
```diff
@@ -10,6 +10,7 @@ applications:
   env:
     RAILS_ENV: production
     RACK_ENV: production
+    CKAN_REDIRECTION_URL: https://ckan.publishing.service.gov.uk/
```
brucebolt: Same as my other comment about the trailing slash.

andreagrandi:
Yay! 👍

PR #684, opened 2019-07-11
Add -u to set for a couple of scripts

cbaines:
So that they fail fast when variables aren't defined.

### DIFF
--- a/scripts/deploy-production.sh
+++ b/scripts/deploy-production.sh
@@ -1,5 +1,5 @@
 #!/bin/bash
 
-set -e
+set -eu
 
 cf zero-downtime-push $CF_APP -f production-manifest.yml --show-app-log=true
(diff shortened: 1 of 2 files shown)


scripts/deploy-production.sh
```diff
@@ -1,5 +1,5 @@
 #!/bin/bash
 
-set -e
+set -eu
 
 cf zero-downtime-push $CF_APP -f production-manifest.yml --show-app-log=true
```
thomasleese: It's still not clear to me why we have this as a variable. We should at least default it to `find-data-beta` I think, it always confuses people when they come to deploy.

cbaines: Yeah, I came across this by trying to run the script, and getting an odd message from `cf`, as this variable was undefined.

PR #686, opened 2019-09-05
Fix the display of licences

kentsanggds:
## What

Fix licences that were not displaying correctly if they have ["..."] in the field.

During an investigation in the data it was discovered that the `licence_custom` field can contain data with or without brackets, pre and post migration data was looked at so both need to be handled.

## Reference

https://trello.com/c/mF4hegw6/1247-dgu-licence-information-including-quotations-in-urls

### DIFF
--- a/app/models/dataset.rb
+++ b/app/models/dataset.rb
@@ -31,7 +31,14 @@ class Dataset
     @released = hash["released"]
     @licence = hash["licence"]
     @licence_other = hash["licence_other"]
-    @licence_custom = hash["licence_custom"]
+    raw_licence_custom = hash["licence_custom"]
+    # left this code in place because extracting it into a method would break tests
+    # as the initialize is called twice in the tests
+    @licence_custom = if raw_licence_custom && raw_licence_custom.match?(/\[.+\]/)
+                        YAML.safe_load(raw_licence_custom).join("<br />")
+                      else
+                        raw_licence_custom
+                      end
     @licence_title = hash["licence_title"]
     @licence_url = hash["licence_url"]
     @licence_code = hash["licence_code"]
--- a/spec/factories/dataset.rb
+++ b/spec/factories/dataset.rb
@@ -63,6 +63,10 @@ FactoryBot.define do
       licence_custom 'Special case'
     end
 
+    trait :with_custom_licence_brackets do
+      licence_custom '["Special case"]'
+    end
+
     initialize_with { Dataset.new(attributes.stringify_keys) }
   end
 end
--- a/spec/features/dataset_show_page_spec.rb
+++ b/spec/features/dataset_show_page_spec.rb
@@ -64,6 +64,24 @@ RSpec.feature 'Dataset page', type: :feature, elasticsearch: true do
       end
     end
 
+    scenario 'Link to custom licence brackets (no title, no URL)' do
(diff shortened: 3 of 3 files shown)


app/models/dataset.rb
```diff
@@ -1,3 +1,5 @@
+require 'yaml'
```
thomasleese: Do we definitely need this? I think this should be automatically `require`d.

app/models/dataset.rb
```diff
@@ -31,7 +33,11 @@ def initialize(hash)
     @released = hash["released"]
     @licence = hash["licence"]
     @licence_other = hash["licence_other"]
-    @licence_custom = hash["licence_custom"]
+
+    if hash["licence_custom"]
+      licence_custom = hash["licence_custom"].match?(/\[.+\]/) ? YAML.safe_load(hash["licence_custom"]).join('<br>') : hash["licence_custom"]
```
thomasleese: It would be more ideomatic Ruby to write this as a memoized method. Something like this:

Assuming we store the value from the hash as `raw_licence_custom`.

```ruby
def licence_custom
  @licence_custom ||= if raw_licence_custom.match?(/\[.+\]/)
                        YAML.safe_load(raw_licence_custom).join("<br />")
                      else
                        raw_licence_custom
                      end
end
```

app/models/dataset.rb
kentsanggds: not sure, will take it out and see

app/models/dataset.rb
kentsanggds: Can we not use `hash["licence_custom"]` directly?

thomasleese: We could.

Personally I wouldn't because it would mean storing `hash` as an instance variable just for this method, and it's ever so slightly less efficient to do the hash lookup multiple times in this method.

kentsanggds: ok, lets try your way

app/models/dataset.rb
```diff
@@ -31,7 +31,12 @@ def initialize(hash)
     @released = hash["released"]
     @licence = hash["licence"]
     @licence_other = hash["licence_other"]
-    @licence_custom = hash["licence_custom"]
+    raw_licence_custom = hash["licence_custom"]
+    @licence_custom ||= if raw_licence_custom && raw_licence_custom.match?(/\[.+\]/)
```
thomasleese: Ah, I actually meant for this to be in a method:

```ruby
def licence_custom
  @licence_custom ||= if @raw_licence_custom && @raw_licence_custom.match?(/\[.+\]/)
                          YAML.safe_load(@raw_licence_custom).join("<br />")
                        else
                          @raw_licence_custom
                        end
end
```

But have `@raw_licence_custom = hash["licence_custom"]` in the constructor.

It's a common pattern in Ruby called memoized methods (https://www.justinweiss.com/articles/4-simple-memoization-patterns-in-ruby-and-one-gem/). It means that the work of parsing the YAML can be pushed to the point when it is actually needed and then cached for the lifetime of the instance. If we left it in the constructor, it would be executed every time the class is constructed, even if the value of `licence_custom` is never requested.

kentsanggds: Tests broke when I tried that, have left a comment to explain why the code is not extracted

PR #687, opened 2019-09-09
Strip whitespace from format for checks on CSV

kentsanggds:
## What

In some cases the format is stored as `CSV ` (additional space) which causes the CSV check to fail so the CSV previews do not show. 

The additional space is usually found in older datasets, the problem is probably due to the `format` field being allowed to be freeform text. In the most recent versions of `ckan` the format is set by the system if the user doesn't enter a `format` value, but it's still possible to enter `CSV `.

Another solution may have been to clean the data but this would also require an update to the ckan publisher to ensure that there are no trailing spaces.

## Reference 

https://trello.com/c/TZj2r5FI/1258-csv-preview-not-available-for-some-historical-transparency-data

### DIFF
--- a/app/models/datafile.rb
+++ b/app/models/datafile.rb
@@ -10,7 +10,7 @@ class Datafile
     @start_date = hash["start_date"]
     @created_at = hash["created_at"]
     @updated_at = hash["updated_at"]
-    @format = hash["format"]
+    @format = hash["format"]&.strip&.upcase
     @size = hash["size"]
     @uuid = hash["uuid"]
   end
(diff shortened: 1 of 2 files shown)


app/models/datafile.rb
```diff
@@ -34,7 +34,7 @@ def html?
   end
 
   def csv?
-    format&.upcase == 'CSV'
+    format[0..2]&.upcase == 'CSV'
```
bevanloon: Is the trailing space going to be a problem for any of the other methods? 
What about trimming `format` in `initialize` or potentially override the accessor to return a trimmed version? I'm not really sure what this class is used for in the broad scheme of things though so I don't know if that would introduce any bugs somewhere else.

kentsanggds: Sure, I could trim it in initialize, I thought that just getting the first 3 chars would be more efficient but am open to changing it although at the same time, I know that only the first 3 chars are needed but this could potentially hide incorrect formats as well.

PR #688, opened 2019-09-16
Update regex for licence_custom

kentsanggds:
## What

Datasets with custom licences were not showing because the licence contained square brackets within the licence. 

The regex to handle the `licence_custom` field is updated to only look for square brackets at the ends.

### DIFF
--- a/spec/factories/dataset.rb
+++ b/spec/factories/dataset.rb
@@ -67,6 +67,10 @@ FactoryBot.define do
       licence_custom '["Special case"]'
     end
 
+    trait :with_custom_licence_brackets_middle do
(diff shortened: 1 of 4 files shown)


spec/factories/dataset.rb
```diff
@@ -67,6 +67,10 @@
       licence_custom '["Special case"]'
     end
 
+    trait :with_custom_licence_brackets_middle do
```
thomasleese: Should we add a test which uses this trait?

kentsanggds: sure, forgot to add

PR #694, opened 2019-10-09
Switch to using Rubocop directly

cbaines:
As this avoids the need for the govuk-lint-ruby wrapper script.


issyl0:
Why is the wrapper script problematic?

cbaines:
> Why is the wrapper script problematic?

My motivation for raising this is that the initial build of https://github.com/alphagov/datagovuk_find/pull/693 failed, as it hit a race condition in the govuk-lint-ruby script.

PR #714, opened 2020-02-07
Clear seen_cookie_message cookie

thomasleese:
This cookie is set automatically by govuk_template but we don't want to set any cookies as we don't have consent.

Updating govuk_template to not set this cookie is going to difficult as the library is deprecated.

[Trello Card](https://trello.com/c/gyV5StET/1747-stop-datagovuk-setting-cookies)

### DIFF
--- a/app/assets/javascripts/govuk.js
+++ b/app/assets/javascripts/govuk.js
@@ -6,6 +6,9 @@ $(document).ready(function () {
   // with role="button" when the space key is pressed.
   GOVUK.shimLinksWithButtonRole.init()
 
+  // Clear cookie set by govuk_template.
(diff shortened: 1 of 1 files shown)


vanitabarrett:
It probably depends how separate you want to keep the code, but you could add it to https://github.com/alphagov/datagovuk_find/blob/master/app/assets/javascripts/govuk.js#L5 as this is already being included ?

thomasleese:
I wasn't sure if that would work because that's put in the `<head>` and [the `govuk_template` code is added later on in the file](https://github.com/alphagov/govuk_template/blob/3bc945d0abe82e6d43a542bc6d8c9088881e1636/source/views/layouts/govuk_template.html.erb#L108).

PR #716, opened 2020-02-10
Delete cookies if present

AlanGabbianelli:
Since we don't ask for consent to set cookies in data.gov.uk and we have
done work to avoid setting them, we also want to delete those that were
previously set.

Trello card: https://trello.com/c/gyV5StET/1747-stop-datagovuk-setting-cookies

### DIFF
--- a/app/assets/javascripts/govuk.js
+++ b/app/assets/javascripts/govuk.js
@@ -8,6 +8,11 @@ $(document).ready(function () {
 
   // Clear cookie set by govuk_template.
   GOVUK.cookie('seen_cookie_message', null);
+  document.cookie = '_ga=;expires=' + new Date() + ';domain=' + window.location.hostname + ';path=/'
+  document.cookie = '_gat=;expires=' + new Date() + ';domain=' + window.location.hostname + ';path=/'
+  document.cookie = '_gid=;expires=' + new Date() + ';domain=' + window.location.hostname + ';path=/'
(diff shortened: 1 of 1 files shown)


app/assets/javascripts/govuk.js
```diff
@@ -8,6 +8,10 @@ $(document).ready(function () {
 
   // Clear cookie set by govuk_template.
   GOVUK.cookie('seen_cookie_message', null);
+  document.cookie = '_ga=;expires=' + new Date() + ';domain=' + window.location.hostname + ';path=/'
+  document.cookie = '_gat=;expires=' + new Date() + ';domain=' + window.location.hostname + ';path=/'
+  document.cookie = '_gid=;expires=' + new Date() + ';domain=' + window.location.hostname + ';path=/'
+  document.cookie = 'dgu_beta_banner_dismissed=;expires=' + new Date() + ';domain=' + window.location.hostname + ';path=/'
```
cbaines: I guess this removes the cookie by setting it to expire now (`new Date()`)... that API is odd.

AlanGabbianelli: Yes, I had a chat with @vanitabarrett about this. I'm not sure if/how we can do this differently. Do you have suggestions?

nickcolley:
I think the cookies starting with _g would be removed if you remove Google Analytics from this website.

vanitabarrett:
@nickcolley I think GA cookies have been disabled already for data.gov.uk - this is a tidy up task to remove the cookies that have may have already been set for users on previous visits.

PR #717, opened 2020-02-19
Use GOVUK cookie banner and consent mechanism in DGU

vanitabarrett:
## What
Use GOVUK's cookie banner and consent mechanism on data.gov.uk

### Cookie Banner
Screenshot 2020-02-21 at 11 17 06

### Cookie Banner confirmation
Screenshot 2020-02-20 at 14 26 11

### Cookie Settings page
Screenshot 2020-02-20 at 14 26 43
Screenshot 2020-02-20 at 14 26 53
Screenshot 2020-02-20 at 14 26 58


app/views/layouts/application.html.erb
```diff
@@ -50,6 +49,8 @@
 
   <div id="global-app-error" class="app-error hidden"></div>
 
+  <%= javascript_include_tag 'application' %>
```
andysellick: Can you explain the reason for moving this line? Seems fine, just curious.

app/views/layouts/_analytics.html.erb
```diff
@@ -1,70 +1,74 @@
...
+        (i[r].q=i[r].q||[]).push(arguments)},i[r].l=1*new Date();a=s.createElement(o),
+                              m=s.getElementsByTagName(o)[0];a.async=1;a.src=g;m.parentNode.insertBefore(a,m)
+      })(window,document,'script','https://www.google-analytics.com/analytics.js','ga')
 
-   $(document).ready(function() {
-     dimensions['dimension11'] = 'find-open-data'
-     ga('send', 'pageview', dimensions)
-     $('[data-ga-event]').on('click', sendAnalyticsEvent)
-   })
+      ga('create', '<%= Rails.application.config.analytics_tracking_id %>', 'auto')
```
andysellick: Looking at the output of the analytics in a browser extension, it looks like it's using a GA property id (tracking id) of `123456789`. This looks like a test value, just wanted to check that when it goes live it'll be set to something else?

vanitabarrett: Yes, this is just a value I've set in the heroku app config as the code expect an environment variable to be set. When it's deployed, it will use the environment variable set in production.

Gemfile
```diff
@@ -11,6 +11,7 @@ gem "faraday"
 gem "faraday_middleware"
 gem "govuk_elements_rails"
 gem "govuk_template"
+gem "govuk_publishing_components", "~> 21.25.0"
```
injms: Is it a worry that Transport could be accidentally used at a later date as that font is bundled into the publishing components?

app/views/layouts/application.html.erb
vanitabarrett: Good question... I *think* I moved it because otherwise the cookie banner Javascript didn't seem to initialise and therefore the cookie banner was just never shown. I assumed this was happening because the JS was being included before the cookie banner.

Gemfile
vanitabarrett: I would say it's not a concern now for DGU right now, perhaps something to consider if we ever want to open up the gem to wider use though. 

I'm not sure on the plan for data.gov.uk, but it looks like [the font is allowed on gov.uk and *.service.gov.uk](https://designnotes.blog.gov.uk/2015/03/11/can-i-use-the-gov-uk-fonts/), so it's possible that could be extended to DGU if we look at bringing it in line visually with gov.uk in the future.

PR #721, opened 2020-02-26
Fix licence url in dataset metadata

deborahchua:
If the dataset `licence_url` and `licence_name` are both `null`,
this invalidates the schema and returns the error
`One of name or url must be provided.` This means the datasets
might not appear as rich results in Google's Dataset Search tool
(https://datasetsearch.research.google.com/)

If the `licence_url` isn't specified  we can set it to the licence
section of the dataset page.

Before:
Screenshot 2020-02-26 at 15 02 07

After:
Screenshot 2020-02-26 at 15 02 53

Trello card: https://trello.com/c/fPAehpfT/1764-5-fix-datagovuk-dataset-markup-errors-as-reported-by-google-search-console

### DIFF
--- a/app/helpers/datasets_helper.rb
+++ b/app/helpers/datasets_helper.rb
@@ -58,7 +58,7 @@ module DatasetsHelper
         "@type": "CreativeWork",
         name: dataset.licence_title,
         text: dataset.licence_custom,
-        url: dataset.licence_url,
+        url: licence_url(dataset),
       },
       dateModified: dataset.public_updated_at,
     }
@@ -83,6 +83,10 @@ module DatasetsHelper
       )
     end
     files
+  end
+
+  def licence_url(dataset)
+    dataset.licence_url.presence || "#{request.protocol}#{request.host_with_port}#{request.fullpath}#license-info"
   end
 
   def contact_information_exists?(dataset)


benjamineskola:
Just checking: should that be `licence_url_exits?` or `licence_url_exists?`?

app/helpers/datasets_helper.rb
```diff
@@ -85,6 +85,11 @@ def metadata_files(dataset)
     files
   end
 
+  def licence_url_exits?(dataset)
+    dataset.licence_url.nil? ? "#{request.protocol}#{request.host_with_port}#{request.fullpath}#license-info" : dataset.licence_url
```
thomasleese:
```suggestion
    dataset.licence_url.presence || "#{request.protocol}#{request.host_with_port}#{request.fullpath}#license-info"
```

This is [a nice little method provided by Rails](https://apidock.com/rails/Object/presence) to avoid needing to reference the `dataset.licence_url` variable twice.

app/helpers/datasets_helper.rb
```diff
@@ -85,6 +85,11 @@ def metadata_files(dataset)
     files
   end
 
+  def licence_url_exits?(dataset)
```
thomasleese:
```suggestion
  def licence_url(dataset)
```

I _think_ this might be a typo of `exists`, but also I think it's better to avoid using the `exists?` part in this name because it implies the return value will be a `true` or `false` rather than the value itself.

app/helpers/datasets_helper.rb
issyl0: This line is too long to read comfortably, I think it would be better as a full `if`.

app/helpers/datasets_helper.rb
deborahchua: Thanks @thomasleese that's a good point! Also indeed was a typo @benjamineskola 😬

app/helpers/datasets_helper.rb
deborahchua: Ah nice thanks @thomasleese ! I've updated this now 👍

PR #723, opened 2020-02-27
2ND Line: highlighting problem with CSRF authentication in support form

GDSNewt:
The recent move to disable session cookies (https://github.com/alphagov/datagovuk_find/commit/44156183dbdc7ec0e461241edc557601b1b9e235) seems to have broken the support forms.

Currently, clicking submit takes the user to the following error...

Screenshot 2020-02-27 at 12 22 51
Screenshot 2020-02-27 at 12 23 08

This is because the server compares the submitted authenticity_token to the value associated with the user’s session. So, without cookies enabled this will not work, since there is no associated session.

No one picked up on this happening as the tests still passed, this is because forgery protection is set to false by default in test and development environments. Toggling this setting to true results in all of the tests in _support_spec.rb_ failing. I have left it on true in order to alert us should this happen again. I don't think it will affect performance with regards to running the tests all that much. 

To restore functionality in the forms I have restored the old settings for session cookies.

### DIFF
--- a/config/environments/test.rb
+++ b/config/environments/test.rb
@@ -25,8 +25,7 @@ Rails.application.configure do
   # Render exceptions instead of raising them
   config.action_dispatch.show_exceptions = true
 
-  # Disable request forgery protection in test environment.
-  config.action_controller.allow_forgery_protection = false
+  # Disable caching in test environment.
   config.action_mailer.perform_caching = false
 
   # Tell Action Mailer not to deliver emails to the real world.
@@ -39,4 +38,7 @@ Rails.application.configure do
 
   # Raises error for missing translations
   config.action_view.raise_on_missing_translations = true
+
+  # The support forms require CSRF authentication, this should be checked in the tests
+  config.action_controller.allow_forgery_protection = true  
 end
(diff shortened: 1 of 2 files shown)


vanitabarrett:
I don't think this will meet the cookie consent guidelines
We don't have a mechanism for users to opt in or out of server-side cookies yet

issyl0:
Whether we're doing this or not, can we have some of the context that's in the PR in the commit message please @GDSNewt? And re-word the first line of the commit message to be under 70 characters as per the [git style guide](https://gds-way.cloudapps.digital/standards/source-code.html#commit-messages).

GDSNewt:
> Whether we're doing this or not, can we have some of the context that's in the PR in the commit message please @GDSNewt? And re-word the first line of the commit message to be under 70 characters as per the [git style guide](https://gds-way.cloudapps.digital/standards/source-code.html#commit-messages).

:shipit: 👍

nicholsj:
> I don't think this will meet the cookie consent guidelines
> We don't have a mechanism for users to opt in or out of server-side cookies yet

These cookies are needed for site functionality and to ensure form submissions are secure. Without them, users can't submit a support request. Therefore they come under the “strictly necessary” or essential category, and so don’t require explicit user consent.

PR #728, opened 2020-03-13
Make the 'Sign in' link on the datasets page go to CKAN signin page

issyl0:
- Previously, this linked to for example
  https://data.gov.uk/dataset/edit/research_and_development_tax_credits
  which led to CKAN giving an error because users must be logged in before
  editing a dataset:

```
403 Forbidden
Access was denied to this resource.
User u'' not authorised to edit research_and_development_tax_credits
```

- The button text says "Sign in", so it's reasonable for users to expect
  to be taken to a signin page and _from there_ find their dataset within
  CKAN.
- It's a bit of a shame we can't do any follow-through after users have
  signed into get them to the `/edit/` page, but that's out of scope of
  this card. If we get reports from users that they want that feature,
  then we can think about it.
- Even though DGU Find/Publish/CKAN don't get much use in Integration
  and Staging, I've used `Plek` instead of hardcoding
  `ckan.publishing.service.gov.uk` for ease of testing on deploy.

https://trello.com/c/v7V8YRvd/1825-3-redirect-non-logged-in-publishers-to-the-sign-up-in-page-rather-than-returning-a-403


app/views/datasets/show.html.erb
```diff
@@ -75,7 +75,7 @@
             <div role="note" aria-label="Publisher information" class="panel panel-border-narrow text">
               <p><%= t('datasets.publishers.information') %></p>
             </div>
-            <p><%= link_to t('datasets.publishers.button'), "/dataset/edit/#{@dataset.legacy_name}", role: 'button', class: 'button', rel: %w(nofollow) %></p>
+            <p><%= link_to t('datasets.publishers.button'), "#{Plek.find("ckan")}/user/login", role: 'button', class: 'button', rel: %w(nofollow) %></p>
```
issyl0: Wait a second... Find is hosted on PaaS, isn't it, so `Plek` won't work.

issyl0: Oh well, `ckan.publishing.service.gov.uk` it is because Find doesn't have a non-production.

cbaines: At least my interpretation of this was that the change was intended for only users who aren't logged in to CKAN, whereas this will affect all users, sending them to the login page, rather than the page for the dataset they want to edit?

issyl0: That sounds possible. If I had a CKAN account I could test what currently happens when users are logged in and click that button... 👀

issyl0: The button text is still "Sign in" whether or not you're signed in. 🤔I'll go back to the drawing board on this!

cbaines: At least from the planning meeting, this sounded more like "handle unauthenticated requests for editing datasets in CKAN by redirecting to the login page".

issyl0: Yep, totally. Having re-read the card, it should be done in CKAN. I think the "Sign in" button not linking to the right page was a red herring.

PR #765, opened 2020-06-12
Add dockerfile

kentsanggds:
## What

Add a `Dockerfile` and `setup-docker.sh file` so that Find can be used on the docker-ckan dev stack.

Also fix kaminari version to 1.1.1 as the latest version is incompatible with the elasticsearch gem.

## Reference 

https://trello.com/c/1RCHolm4/180-add-find-to-docker-stack


Gemfile
```diff
@@ -15,7 +15,7 @@ gem "govuk_publishing_components", "~> 21.34.0"
 gem "htmlentities", "~> 4.3"
 gem "jbuilder", "~> 2.5"
 gem "jquery-rails"
-gem "kaminari", "~> 1.2.1"
+gem "kaminari", "= 1.1.1"
```
AlanGabbianelli, line 18:
```suggestion
gem "kaminari", "< 1.2"
```
This will allow patches and bugfixes (e.g. version 1.1.2)

kentsanggds: ok sure

PR #768, opened 2020-06-18
Add invisible span for CSV preview links

owenatgov:
## What
This PR:

1. adds invisible spans to the "preview" links for CSV resources in dataset resource tables.
2. applies the invisible text only approach to a previous similar issue applied to data links <https://github.com/alphagov/datagovuk_find/pull/512>
3. removes the "go to site" link on HTML resources.

## Why
This is to comply with WCAG accessibility needs based on an audit last year. Specifically this makes the preview link now make sense to screen reader users out of context. The 2 links that were triggering this failure where the "go to site" link for HTML resources and "preview" links for CSV resources. The latter required solving, however the former may be redundant as this link is identical to the data link in the first table cell for HTML resources. It therefore makes more sense to simply remove it as the link is a suitable source for the resource itself.

**Card:** https://trello.com/c/WJFeCU9B/184-fix-non-descriptive-or-duplicate-links-from-datasets-high

### DIFF
--- a/app/views/datasets/_datafile_table.html.erb
+++ b/app/views/datasets/_datafile_table.html.erb
@@ -14,10 +14,10 @@
                 'ga-event' => "download",
                 'ga-format' => "CSV",
                 'ga-publisher' => @dataset.organisation.name
-              },
-              'aria-label': "Download datafile #{datafile.name} format: #{format_of(datafile)} dataset: #{@dataset.title}" do %>
-            <%= (datafile.name ? datafile.name : 'Data Link') %>
-            <span class='visually-hidden'>Download datafile '<%= datafile.name %>', Format: <%= format_of(datafile) %>, Dataset: <%= @dataset.title %></span>
+              } do %>
+            <span class="visually-hidden">Download </span>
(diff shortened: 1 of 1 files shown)


risicle:
This WFM on my docker stack FWIW.

maxgds:
I don't seem to have permissions to review this, so here's a "please change" review in a comment.

I know you're following a previous example, but it's not an ideal example. We should choose either the aria approach or the hidden text approach, not both. Screen readers will usually take the aria label to read by default, so the hidden text is then pointless. On a label aria is probably the more reliable approach, but hidden text is valid too and for a link I'd personally favour hidden text. https://accessible360.com/accessible360-blog/use-aria-label-screen-reader-text/

If you decide to stick with the hidden text approach then you want to rethink it slightly, at the moment you'll get repetition as it will read "Preview preview CSV". See a good hidden text example here: https://design-system.service.gov.uk/components/summary-list/ (search for the visually-hidden class).

It's probably worth deciding which approach to take and revisiting the example you copied to bring it in line also.

PR #769, opened 2020-06-19
Set dataset contact links to use text instead of their URLs

owenatgov:
## What
Amends the setup for dataset contact details so that links are no longer rendered in their literal form but use either the specified text for that contact detail or a generic label. Additionally, this adds the `foi_email_is_email` helper and tidies up the contact partial.

## Why
This is to comply with WCAG regulations around descriptive links, specifically in relation to a previous violation following a DAC audit. See page 28 https://drive.google.com/file/d/1r32grtcw7PFNbJnMsXFCM8EPX_GaD753/view

## Testing

On data.gov.uk dataset view ([example](https://data.gov.uk/dataset/115cb6bc-e29e-4a94-95f5-8461db105b20/mortality-target-monitoring)), contacts set for that dataset will now render as a link to the specified email with the link text being the name of the contact

### Before
Screenshot 2020-07-03 at 09 39 06

### After
Screenshot 2020-07-03 at 09 38 24

**Card:** https://trello.com/c/fMi12rao/187-no-descriptive-link-text-on-some-links-low

### DIFF
--- a/app/views/datasets/_additional_info.html.erb
+++ b/app/views/datasets/_additional_info.html.erb
@@ -72,9 +72,9 @@
             <% if i['harvest_object_id'] %>
               <dt><%= t('.inspire_gemini_record') %></dt>
               <dd>
-                <%= link_to t('.xml'), "/api/2/rest/harvestobject/#{ i['harvest_object_id'] }/xml", class: 'govuk_link' %>
+                <%= link_to t('.xml'), "/api/2/rest/harvestobject/#{ i['harvest_object_id'] }/xml", class: 'govuk-link' %>
                 <br />
-                <%= link_to t('.html'), "/api/2/rest/harvestobject/#{ i['harvest_object_id'] }/html", class: 'govuk_link' %>
+                <%= link_to t('.html'), "/api/2/rest/harvestobject/#{ i['harvest_object_id'] }/html", class: 'govuk-link' %>
               </dd>
             <% end %>
           </dl>
--- a/app/views/datasets/_additional_info.html.erb
+++ b/app/views/datasets/_additional_info.html.erb
@@ -71,11 +71,8 @@
             <% end %>
             <% if i['harvest_object_id'] %>
               <dt><%= t('.inspire_gemini_record') %></dt>
-              <dd>
-                <%= link_to t('.xml'), "/api/2/rest/harvestobject/#{ i['harvest_object_id'] }/xml", class: 'govuk-link' %>
-                <br />
-                <%= link_to t('.html'), "/api/2/rest/harvestobject/#{ i['harvest_object_id'] }/html", class: 'govuk-link' %>
-              </dd>
+              <dd><%= link_to t('.xml'), "/api/2/rest/harvestobject/#{ i['harvest_object_id'] }/xml", class: 'govuk-link' %></dd>
+              <dd><%= link_to t('.html'), "/api/2/rest/harvestobject/#{ i['harvest_object_id'] }/html", class: 'govuk-link' %></dd>
             <% end %>
           </dl>
         </div>
--- a/app/views/datasets/_contact.html.erb
+++ b/app/views/datasets/_contact.html.erb
@@ -52,7 +52,7 @@
         <%= content_tag(:span, foi_mail) if foi_mail.present? %>
         <% if foi_web_address_for(dataset).present? %>
           <%= content_tag(:span) do %>
-            <%= link_to("Freedom of information requests for #{@dataset.organisation.title}", foi_web_address_for(dataset), ga_data) %>
+            <%= link_to("Freedom of information requests for this dataset", foi_web_address_for(dataset), ga_data) %>
           <% end %>
         <% end %>
 
--- a/app/views/datasets/_contact.html.erb
+++ b/app/views/datasets/_contact.html.erb
@@ -1,6 +1,6 @@
 <%
   default_name = "Contanct #{@dataset.organisation.title} regarding this dataset"
-  ga_data = {'data-ga-event': 'contact', 'data-ga-publisher': @dataset.organisation.name}
+  link_attrs = {'data-ga-event': 'contact', 'data-ga-publisher': @dataset.organisation.name, 'class': 'govuk_link'}
 
   if contact_name_for(dataset).present?
     email_name = contact_name_for(dataset)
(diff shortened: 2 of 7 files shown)


app/views/datasets/_contact.html.erb
```diff
@@ -1,8 +1,29 @@
 <%
+  default_name = "Contanct #{@dataset.organisation.title} regarding this dataset"
```
nicholsj, line 2: Typo

app/views/datasets/_contact.html.erb
```diff
@@ -1,8 +1,29 @@
 <%
+  default_name = "Contact #{@dataset.organisation.title} regarding this dataset"
+  link_attrs = {'data-ga-event': 'contact', 'data-ga-publisher': @dataset.organisation.name, 'class': 'govuk_link'}
+
+  if contact_name_for(dataset).present?
```
danacotoran, line 5: I'm a ruby newbie so ignore me if I'm totally off base on this one, but I was thinking using a ternary operator might be more elegant here?
I.E something like
`email_name = contact_name_for(dataset).present? ? contact_name_for(dataset) : default_name`

I see why you would've gone with the if/else/end option for readability though

app/views/datasets/_contact.html.erb
```diff
@@ -1,9 +1,11 @@
...
-  if contact_email_is_email?(dataset)
-    mail = mail_to(contact_email_for(dataset), nil, {'data-ga-event': 'contact', 'data-ga-publisher': @dataset.organisation.name, class: 'govuk_link'})
-  else
-    mail = link_to(contact_email_for(dataset), contact_email_for(dataset), {'data-ga-event': 'contact', 'data-ga-publisher': @dataset.organisation.name, class: 'govuk_link'})
-  end
+  default_name = "Contact #{@dataset.organisation.title} regarding this dataset"
+  link_attrs = {'data-ga-event': 'contact', 'data-ga-publisher': @dataset.organisation.name, 'class': 'govuk_link'}
+
+  email_name = contact_name_for(dataset).present? ? contact_name_for(dataset) : default_name
+  foi_name = foi_name_for(dataset).present? ? foi_name_for(dataset) : default_name
```
thomasleese, line 6:
```suggestion
  foi_name = foi_name_for(dataset).presence || default_name
```

I believe you can use the `presence` method here to simplify the code: https://apidock.com/rails/Object/presence

app/views/datasets/_contact.html.erb
```diff
@@ -1,9 +1,11 @@
 <%
-  if contact_email_is_email?(dataset)
-    mail = mail_to(contact_email_for(dataset), nil, {'data-ga-event': 'contact', 'data-ga-publisher': @dataset.organisation.name, class: 'govuk_link'})
-  else
-    mail = link_to(contact_email_for(dataset), contact_email_for(dataset), {'data-ga-event': 'contact', 'data-ga-publisher': @dataset.organisation.name, class: 'govuk_link'})
-  end
+  default_name = "Contact #{@dataset.organisation.title} regarding this dataset"
```
maxgds, line 2: Could you change the variable name to something like default_message to better represent what it contains? 

I'm also wondering about whether it should be bare text in the code - generally we try to put strings into the translation files, and I know it's possible to pass a parameter in too, though I can't remember the syntax. Don't worry about it too much if there are other examples of bare strings, though. Likewise for examples further down this PR.

app/views/datasets/_additional_info.html.erb
```diff
@@ -72,9 +72,9 @@
             <% if i['harvest_object_id'] %>
               <dt><%= t('.inspire_gemini_record') %></dt>
               <dd>
-                <%= link_to t('.xml'), "/api/2/rest/harvestobject/#{ i['harvest_object_id'] }/xml", class: 'govuk_link' %>
+                <%= link_to t('.xml'), "/api/2/rest/harvestobject/#{ i['harvest_object_id'] }/xml", class: 'govuk-link' %>
```
alex-ju, line 75: Non-blocker: I'm aware this wasn't introduced by this change, but I would suggest we improve the semantics too by using two `dd`s instead of one split using a `br`
```erb
<dt><%= t('.inspire_gemini_record') %></dt>
<dd><%= link_to t('.xml'), "/api/2/rest/harvestobject/#{ i['harvest_object_id'] }/xml", class: 'govuk-link' %></dd>
<dd><%= link_to t('.html'), "/api/2/rest/harvestobject/#{ i['harvest_object_id'] }/html", class: 'govuk-link' %></dd>
```
It may require some additional styling.

app/views/datasets/_contact.html.erb
alex-ju, line 2: Would be great is we move this hardcoded copy into `locales`/translation file – they allow for parameters to be passed in, to deal with title.

app/views/datasets/_contact.html.erb
```diff
@@ -29,11 +30,10 @@
       </h3>
 
       <p>
-        <%= content_tag(:span, foi_name_for(dataset)) if foi_name_for(dataset).present? %>
-        <%= content_tag(:span, mail_to(foi_email_for(dataset), nil, {class: 'ga-contact'})) if foi_email_for(dataset).present? %>
+        <%= content_tag(:span, foi_mail) if foi_mail.present? %>
         <% if foi_web_address_for(dataset).present? %>
           <%= content_tag(:span) do %>
-            <%= link_to(foi_web_address_for(dataset), foi_web_address_for(dataset), { class: 'ga-contact'} ) %>
+            <%= link_to("Freedom of information requests for this dataset", foi_web_address_for(dataset), link_attrs) %>
```
alex-ju, line 36: Same suggestion here regarding hardcoded text.

spec/helpers/datasets_helper_spec.rb
```diff
@@ -22,4 +22,16 @@
       expect(helper.contact_email_is_email?(dataset)).to be_falsey
     end
   end
+
+  describe "#foi_email_is_email?" do
+    it "returns true when the email is valid" do
```
maxgds, line 27: Nitpick: I'd be a bit more explicit with the description. You're just checking for an @ so a@b would pass your test but it's not a valid email address. Something like "returns true when the email address contains a @"

kentsanggds:
I've tested the changes on my local docker ckan stack and the links are showing correctly, so once the minor changes to the code are in place it's good to go.

PR #778, opened 2020-07-07
Update markup on the data.gov.uk hompage

owenatgov:
## What
Update the markup, classnames and styling on the homepage view and associated components (search and search topics). There should be no discernible difference between this PR and production.

## Why

**Card:** https://trello.com/c/MHM8t3z8/231-update-markup-on-find-homepage

### DIFF
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -23,6 +23,8 @@ $govuk-use-legacy-palette: false;
 @import 'govuk_publishing_components/components/_layout-header';
 @import 'govuk_publishing_components/components/_layout-footer';
 @import 'govuk_publishing_components/components/_title';
+@import 'govuk_publishing_components/components/_search';
+@import 'govuk_publishing_components/components/_lead-paragraph';
 @import 'govuk_publishing_components/components/_notice';
 @import 'govuk_publishing_components/components/_heading';
 @import 'govuk_publishing_components/components/_govspeak';
--- a/app/assets/stylesheets/components.scss
+++ b/app/assets/stylesheets/components.scss
@@ -16,25 +16,19 @@
 }
 
 // Home ==================================================================
-.dgu-home-search {
-  padding-top: 2.5em;
-}
-
 .dgu-topics {
-
-  margin-top: 2em;
-
-  h2 {
-    font-weight: bold;
+  &__heading {
+    margin-bottom: 5px;
   }
+  &__list {
+    columns: 3;
 
-  @media (min-width: 641px) {
     li {
-      min-height: 6em;
+      display: inline-block;
+      width: 100%;
     }
   }
 }
-
 
 // Search box ==================================================================
 .dgu-search-box {
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -22,7 +22,7 @@
   <body class="govuk-template__body">
     <script>document.body.className = ((document.body.className) ? document.body.className + ' js-enabled' : 'js-enabled');</script>
 
-    <a href="#content" class="govuk-skip-link"><%= content_for?(:skip_link_message) ? yield(:skip_link_message) : "Skip to main content" %></a>
+    <%= render "govuk_publishing_components/components/skip_link", {} %>
 
     <%= render "govuk_publishing_components/components/cookie_banner", {
       text: sanitize("We use <a href='/cookies' class='govuk-link'>cookies to collect information</a> about how you use data.gov.uk. We use this information to make the website work as well as possible."),
@@ -49,7 +49,9 @@
       </div>
     </div>
 
-    <%= yield %>
+    <div class="govuk-width-container">
+      <%= yield %>
+    </div>
 
     <%= render "govuk_publishing_components/components/layout_footer", {
       meta: {
--- a/app/views/pages/_search.html.erb
+++ b/app/views/pages/_search.html.erb
@@ -1,20 +1,7 @@
-<div class="grid-row">
-  <div class="column-two-thirds">
-    <h1 class="heading-xlarge">
-      <%= t('.find_gov_data') %>
-    </h1>
-    <p class="lede">
-    <%= t('.lede') %>
-    </p>
-  </div>
-</div>
-<div class="grid-row">
-  <div class="column-two-thirds dgu-search-box">
-    <form action="/search" method="GET">
-      <label for="q" class="visuallyhidden"><%= t('.accessibility.search_box_label') %></label>
-      <input id="q" name="q" type="text" class="form-control dgu-search-box__input" /><!-- Make sure there is no space between the input and button else the button will become detached --><button type="submit" class="dgu-search-box__button">
-        <%= t('.accessibility.search_button_label') %>
-      </button>
-    </form>
-  </div>
+<div class="govuk-grid-row">
+  <form action="/search" method="GET" class="govuk-grid-column-two-thirds">
+    <%= render "govuk_publishing_components/components/search", {
+      label_text: t('.accessibility.search_button_label')
+    } %>
+  </form>
 </div>
--- a/app/views/pages/_topics.html.erb
+++ b/app/views/pages/_topics.html.erb
@@ -1,69 +1,55 @@
-<div class="grid-row dgu-topics">
-  <div class="column-one-third">
-    <div>
-      <ol>
-        <li>
-          <h2><%= link_to t('.business_and_economy_label'), search_path(filters: { topic: 'Business and economy' }), class: 'govuk-link' %></h2>
-          <p><%= t('.business_and_economy_description') %></p>
-        </li>
-        <li>
-          <h2><%= link_to t('.crime_and_justice_label'), search_path(filters: { topic: 'Crime and justice' }), class: 'govuk-link' %></h2>
-          <p><%= t('.crime_and_justice_description') %></p>
-        </li>
-        <li>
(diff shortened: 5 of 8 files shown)


app/views/pages/_search.html.erb
```diff
@@ -1,18 +1,8 @@
...
-</div>
-<div class="grid-row">
-  <div class="column-two-thirds dgu-search-box">
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds dgu-search-box">
     <form action="/search" method="GET">
-      <label for="q" class="visuallyhidden"><%= t('.accessibility.search_box_label') %></label>
-      <input id="q" name="q" type="text" class="form-control dgu-search-box__input" /><!-- Make sure there is no space between the input and button else the button will become detached --><button type="submit" class="dgu-search-box__button">
+      <label for="search" class="govuk-visually-hidden"><%= t('.accessibility.search_box_label') %></label>
+      <input id="search" name="search" type="text" class="form-control dgu-search-box__input" /><!-- Make sure there is no space between the input and button else the button will become detached --><button type="submit" class="dgu-search-box__button">
```
andysellick: Could you use the search component here?

andysellick: If this comment is essential, suggest it be done as a ruby comment, otherwise it's visible in the public source (although this seems like something that could be fixed).

app/assets/stylesheets/components.scss
```diff
@@ -16,26 +16,12 @@
...
-
-  h2 {
-    font-weight: bold;
-  }
-
-  @media (min-width: 641px) {
-    li {
-      min-height: 6em;
-    }
+  &__heading {
```
andysellick, line 20: If this is only going to be one rule could it just be written as `.dgu-topics__heading`?

app/views/pages/_topics.html.erb
```diff
@@ -1,69 +1,64 @@
...
-          <h2><%= link_to t('.defence_label'), search_path(filters: { topic: 'Defence' }), class: 'govuk-link' %></h2>
-          <p><%= t('.defence_description') %></p>
-        </li>
-        <li>
-          <h2><%= link_to t('.education_label'), search_path(filters: { topic: 'Education' }), class: 'govuk-link' %></h2>
-          <p><%= t('.education_description') %></p>
-        </li>
-      </ol>
-    </div>
+<div class="govuk-grid-row dgu-topics">
```
andysellick, line 1: Is the `dgu-topics` class used? I can't find it in the stylesheet.

app/assets/stylesheets/application.scss
```diff
@@ -23,6 +23,7 @@ $govuk-use-legacy-palette: false;
 @import 'govuk_publishing_components/components/_layout-header';
 @import 'govuk_publishing_components/components/_layout-footer';
 @import 'govuk_publishing_components/components/_title';
+@import 'govuk_publishing_components/components/_search';
```
andysellick, line 26: Isn't this change already in https://github.com/alphagov/datagovuk_find/pull/779/files?

owenatgov: Indeed it is. Because both these PR's required very similar changes, I had a plan to merge one and then rebase the other. Whichever PR get's merged first is the one that introduced the change. I wanted to avoid making a separate PR for just this, mainly if I'm honest because I'm finding it difficult to get review on these datagovuk PRs.

app/views/pages/_search.html.erb
owenatgov: I do update this later on. Do you think it's worth me interactive rebasing back through this?

andysellick: Maybe? But only if it's not a pain. Sorry, I didn't notice it later on.

owenatgov: I think it makes more sense for the commit history to be healthy so i'll sort it out.

app/views/layouts/application.html.erb
```diff
@@ -22,7 +22,7 @@
   <body class="govuk-template__body">
     <script>document.body.className = ((document.body.className) ? document.body.className + ' js-enabled' : 'js-enabled');</script>
 
-    <a href="#content" class="govuk-skip-link"><%= content_for?(:skip_link_message) ? yield(:skip_link_message) : "Skip to main content" %></a>
+    <a href="#main-content" class="govuk-skip-link"><%= content_for?(:skip_link_message) ? yield(:skip_link_message) : "Skip to main content" %></a>
```
injms, line 25: Is it worth using a [skip link component](https://components.publishing.service.gov.uk/component-guide/skip_link) here?

app/views/pages/_search.html.erb
```diff
@@ -1,20 +1,9 @@
...
-    <p class="lede">
-    <%= t('.lede') %>
-    </p>
-  </div>
-</div>
-<div class="grid-row">
-  <div class="column-two-thirds dgu-search-box">
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds dgu-search-box">
     <form action="/search" method="GET">
```
injms, line 3: Slightly picky - you could remove a `div` element by adding `class="govuk-grid-column-two-thirds dgu-search-box"` to this form element.
```suggestion
    <form action="/search" method="GET" class="govuk-grid-column-two-thirds dgu-search-box">
```

app/views/pages/_topics.html.erb
```diff
@@ -1,69 +1,64 @@
...
-        <li>
-          <h2><%= link_to t('.education_label'), search_path(filters: { topic: 'Education' }), class: 'govuk-link' %></h2>
-          <p><%= t('.education_description') %></p>
-        </li>
-      </ol>
-    </div>
+<div class="govuk-grid-row">
+  <h2 class="govuk-visually-hidden"><%= t('.topics_heading') %></h2>
+  <div class="govuk-grid-column-one-third">
+    <ul class="govuk-list">
```
injms, line 4: This isn't a blocker - the (recently added) [list component](https://components.publishing.service.gov.uk/component-guide/list) could be used here.

app/views/pages/_topics.html.erb
```diff
@@ -1,69 +1,64 @@
...
+        <p class="govuk-body"><%= t('.education_description') %></p>
+      </li>
+    </ul>
   </div>
 
-  <div class="column-one-third">
-    <div>
-      <ol>
+  <div class="govuk-grid-column-one-third">
+      <ul>
```
injms, line 25: Does this unordered list need the `govuk-list` class?

```suggestion
      <ul class="govuk-list">
```

app/views/pages/_topics.html.erb
```diff
@@ -1,69 +1,64 @@
...
+          <p class="govuk-body"><%= t('.health_description') %></p>
         </li>
-      </ol>
-    </div>
+      </ul>
   </div>
-  <div class="column-one-third">
-    <div>
+  <div class="govuk-grid-column-one-third">
       <ol>
```
injms, line 45: The previous two columns have unordered lists rather than ordered lists - should this column be the same?
```suggestion
      <ul class="govuk-list">
```

app/views/pages/_topics.html.erb
```diff
@@ -1,69 +1,64 @@
...
-          <p><%= t('.defence_description') %></p>
-        </li>
-        <li>
-          <h2><%= link_to t('.education_label'), search_path(filters: { topic: 'Education' }), class: 'govuk-link' %></h2>
-          <p><%= t('.education_description') %></p>
-        </li>
-      </ol>
-    </div>
+<div class="govuk-grid-row">
+  <h2 class="govuk-visually-hidden"><%= t('.topics_heading') %></h2>
```
injms, line 2: Larger comment on the three columns list - the way this is coded means that there are three lists disguised as one list.

This means that people who use screen readers (and similar assistive technologies) will have to navigate three lists - but for people who use the site visually is looks like one list.

Ideally this would be one list with twelve items, rather than three lists with four items each.

I don't know what DGU's browser support is, but assuming it's the same as GOV.UK Publishing we could look at doing this with [CSS columns](https://caniuse.com/#search=columns) with a single column fallback.

injms, line 2: Larger comment on the three columns list - the way this is coded means that there are three lists disguised as one list.

This means that people who use screen readers (and similar assistive technologies) will have to navigate three lists - but for people who use the site visually is looks like one list.

Ideally this would be one list with twelve items, rather than three lists with four items each.

I don't know what DGU's browser support is, but assuming it's the same as GOV.UK Publishing we could look at doing this with [CSS columns](https://caniuse.com/#search=columns) with a single column fallback.

app/views/pages/_topics.html.erb
owenatgov: Discussed on slack. Because there's a need to add column styling for the `ul` I'm going to build this using markup and classes.

owenatgov:
Rubocop is failing but the vast majority of these offences are going to be handled by this PR https://github.com/alphagov/datagovuk_find/pull/787

app/assets/stylesheets/components.scss
```diff
@@ -16,26 +16,19 @@
...
+    margin-bottom: 5px;
   }
+  &__list {
+    columns: 3;
 
-  @media (min-width: 641px) {
     li {
-      min-height: 6em;
+      display: inline-block;
```
injms, line 27: To help with browsers that don't support columns, could we add a width?
```suggestion
      display: inline-block;
      width: 100%;
```

Before: 
image

After:
image

PR #779, opened 2020-07-08
Update search markup

owenatgov:
## What
Update the markup, classnames and styling on the search view and associated components (filters and sort). There should be no discernible difference between this PR and production.

## Why

**Card:** https://trello.com/c/EbYBZqRv/232-update-markup-on-find-search-page

### DIFF
--- a/app/assets/javascripts/govuk.js
+++ b/app/assets/javascripts/govuk.js
@@ -15,33 +15,25 @@ $(document).ready(function () {
   var showHide = new ShowHide()
   showHide.init()
 
-  if (document.getElementById('publisher')) {
-    accessibleAutocomplete.enhanceSelectElement({
-      selectElement: document.querySelector('#publisher'),
-      showAllValues: true,
-      preserveNullOptions: true
-    })
+  var autocompleteSelects = document.getElementsByClassName('js-accessible-autocomplete')
+
+  if (autocompleteSelects.length > 0) {
+    for (var i = 0; i < autocompleteSelects.length; i += 1) {
+      accessibleAutocomplete.enhanceSelectElement({
+        selectElement: autocompleteSelects[i],
+        showAllValues: true,
+        preserveNullOptions: true
+      })
+    }
   }
 
-  if (document.getElementById('format')) {
-    accessibleAutocomplete.enhanceSelectElement({
-      selectElement: document.querySelector('#format'),
-      showAllValues: true,
-      preserveNullOptions: true
+  var sortDatasets = document.getElementById('sort-datasets')
+
+  if (sortDatasets) {
+    sortDatasets.addEventListener('change', function(event) {
+      event.target.form.submit()
     })
   }
-
-  if(document.getElementById('topic')) {
-    accessibleAutocomplete.enhanceSelectElement({
-      selectElement: document.querySelector('#topic'),
-      showAllValues: true,
-      preserveNullOptions: true
-    })
-  }
-
-  document.getElementById('sort-datasets').addEventListener('change', function(event) {
-    event.target.form.submit()
-  })
 
   new FoldableText('.js-summary', 200)
     .init()
--- a/app/assets/javascripts/govuk.js
+++ b/app/assets/javascripts/govuk.js
@@ -39,6 +39,10 @@ $(document).ready(function () {
     })
   }
 
+  document.getElementById('sort-datasets').addEventListener('change', function(event) {
+    event.target.form.submit()
+  })
+
   new FoldableText('.js-summary', 200)
     .init()
 
@@ -69,7 +73,7 @@ ShowHide.prototype = {
     event.preventDefault()
 
     var openCloseAllControl = $(event.target)
-    var dataLinks = $(this.selector);
+    var dataLinks = $(this.selector)
     var dataLinkContent = dataLinks.find(this.contentSelector)
     var dataLinkExpand = dataLinks.find(this.expandSelector)
     var allOpen = openCloseAllControl.data('allOpen')
@@ -156,7 +160,7 @@ LimitDatasets.prototype.toggle = function (event) {
   if (folded === 'folded') {
     $target.text('Show less')
     this.moreFiles.show()
-    $(this.moreFiles[0]).attr('tabindex', -1).focus();
+    $(this.moreFiles[0]).attr('tabindex', -1).focus()
     $target.data('folded', 'unfolded')
   } else {
     $target.text('Show more')
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -25,6 +25,8 @@ $govuk-use-legacy-palette: false;
 @import 'govuk_publishing_components/components/_title';
 @import 'govuk_publishing_components/components/_search';
 @import 'govuk_publishing_components/components/_lead-paragraph';
+@import 'govuk_publishing_components/components/_select';
+@import 'govuk_publishing_components/components/_checkboxes';
 @import 'govuk_publishing_components/components/_notice';
 @import 'govuk_publishing_components/components/_heading';
 @import 'govuk_publishing_components/components/_govspeak';
--- a/app/views/search/_filters.html.erb
+++ b/app/views/search/_filters.html.erb
@@ -1,32 +1,32 @@
-<h2 class="heading-medium"> <%= t('.filter_by') %></h2>
+<h2 class="heading-medium"> <%= t(".filter_by") %></h2>
 
-<div class="form-group">
-  <%= label_tag 'publisher', t('.filter_by_publisher'), class: 'form-label' %>
-  <%= select_tag('filters[publisher]', options_for_select(dataset_publishers_for_select, selected_publisher), id: 'publisher', class: 'form-control dgu-filters__filter', include_blank: true) %>
+<div class="govuk-form-group">
+  <%= label_tag "publisher", t(".filter_by_publisher"), class: "govuk-label" %>
+  <%= select_tag("filters[publisher]", options_for_select(dataset_publishers_for_select, selected_publisher), id: "publisher", class: "govuk-select dgu-filters__filter", include_blank: true) %>
 </div>
 
-<div class="form-group">
-  <%= label_tag 'topic', t('.filter_by_topic'), class: 'form-label' %>
-  <%= select_tag('filters[topic]', options_for_select(dataset_topics_for_select, selected_topic), id: 'topic', class: 'form-control dgu-filters__filter', include_blank: true) %>
+<div class="govuk-form-group">
+  <%= label_tag "topic", t(".filter_by_topic"), class: "govuk-label" %>
+  <%= select_tag("filters[topic]", options_for_select(dataset_topics_for_select, selected_topic), id: "topic", class: "govuk-select dgu-filters__filter", include_blank: true) %>
 </div>
 
-<div class="form-group">
-  <%= label_tag 'format', t('.filter_by_format'), class: 'form-label' %>
-  <%= select_tag('filters[format]', options_for_select(datafile_formats_for_select, selected_format), id: 'format', class: 'form-control dgu-filters__filter', include_blank: true) %>
+<div class="govuk-form-group">
+  <%= label_tag "format", t(".filter_by_format"), class: "govuk-label" %>
+  <%= select_tag("filters[format]", options_for_select(datafile_formats_for_select, selected_format), id: "format", class: "govuk-select dgu-filters__filter", include_blank: true) %>
 </div>
 
-<div class="form-group">
-  <div class="multiple-choice">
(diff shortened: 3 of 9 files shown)


app/assets/stylesheets/govuk-overrides.scss
```diff
@@ -86,3 +86,7 @@ button.button.secondary {
     background-image: image-url("separator-2x.png");
   }
 }
+
+.gem-c-search--on-white .gem-c-search__submit {
```
andysellick, line 90: What's the reason for this override?

app/assets/stylesheets/application.scss
```diff
@@ -23,6 +23,7 @@ $govuk-use-legacy-palette: false;
 @import 'govuk_publishing_components/components/_layout-header';
 @import 'govuk_publishing_components/components/_layout-footer';
 @import 'govuk_publishing_components/components/_title';
+@import 'govuk_publishing_components/components/_search';
```
andysellick, line 26: This is the stylesheet for the search component, but I can't see the component being called in the code. Is it in a separate commit or already present somewhere?

app/views/search/_filters.html.erb
```diff
@@ -1,32 +1,32 @@
 <h2 class="heading-medium"> <%= t('.filter_by') %></h2>
 
-<div class="form-group">
-  <%= label_tag 'publisher', t('.filter_by_publisher'), class: 'form-label' %>
-  <%= select_tag('filters[publisher]', options_for_select(dataset_publishers_for_select, selected_publisher), id: 'publisher', class: 'form-control dgu-filters__filter', include_blank: true) %>
+<div class="govuk-form-group">
+  <%= label_tag 'publisher', t('.filter_by_publisher'), class: 'govuk-label' %>
+  <%= select_tag('filters[publisher]', options_for_select(dataset_publishers_for_select, selected_publisher), id: 'publisher', class: 'govuk-select dgu-filters__filter', include_blank: true) %>
```
andysellick, line 5: This looks like the markup for the select component, is that right? Why is it written out rather than calling the component? e.g. 

```
<%= render "govuk_publishing_components/components/select", {
  id: "dropdown1",
  label: "My Dropdown",
  options: [
    {
      text: "Option one",
      value: "option1"
    },
  ]
} %>

File: app/views/search/_sort.html.erb [hunk 4]
```diff
@@ -1,11 +1,11 @@
 <div class="dgu-sort">
-  <div class="dgu-sort__apply visuallyhidden">
+  <div class="dgu-sort__apply govuk-visually-hidden">
     <input type="submit" value=<%= t('.accessibility.submit_button') %>/>
   </div>
 
   <div class="dgu-sort__sort-type">
     <label>
-      <span class="visuallyhidden"> <%= t('.accessibility.sort_type_label') %></span>
+      <span class="govuk-visually-hidden"> <%= t('.accessibility.sort_type_label') %></span>
```
andysellick, line 8: This PR deployment isn't working, and I don't have context of the UI, why is this label text being visually hidden?

andysellick, line 8: If you want to visually hide the label, could you put the `govuk-visually-hidden` class on the label itself?

app/assets/stylesheets/govuk-overrides.scss
owenatgov: The button generated by the GUPC instance of the search bar is blue whilst the search bar button across data.gov.uk currently is black. I'm not sure if it makes more sense to just leave it blue, if there's been research done on this etc. The team's in an awkward position where we can't really get design support so I don't want to make any non-trivial changes. If you think I could get away with a blue button I'm happy to chat to the team and scoop this out.

app/views/search/_filters.html.erb
owenatgov: This is an accessible autocomplete instance. I wasn't sure if it was possible to include the select component here and for the AA to still work. I will try this out and get back to you.

app/views/search/_sort.html.erb
owenatgov: It's an accessible label that's sitting next to a search filter. You can see it on this page https://data.gov.uk/search?q= next to the filter dropdown with the selection "best match". I will however make the label change you noted, good spot.

app/assets/stylesheets/application.scss
andysellick: Ah, sorry, never mind then.

app/views/search/_filters.html.erb
andysellick: It probably isn't, so this is probably fine as it is. It wasn't obvious that an AA is applied here, is that done through a specific class? I think it might be worth making this more obvious, perhaps by using a specific `js-` prefixed class for example.

app/assets/stylesheets/govuk-overrides.scss
andysellick: One of the principles of our components is that they're isolated, which means we can have confidence that if it looks correct in the component guide, it looks correct everywhere. If we style components in applications like this, it can lead to problems - for example, if the component was updated to use `govuk-text-colour` as the text colour, in your application the text on the button would suddenly become unreadable. That's not likely to happen, but the rule is there to defend against situations that are more plausible.

The other advantage of not styling components in applications is that we get a consistent look across the site. Is there a reason why DGU has subtle differences like this?

app/views/search/_sort.html.erb
andysellick: I'm still interested to know why the label is hidden. If you look at https://www.gov.uk/search/all?keywords=minister&order=relevance there's a similar 'sort by' option (beneath the result count) and the label is shown.

app/assets/stylesheets/govuk-overrides.scss
owenatgov: I can't dispute this. The question has been asked, will report back.

app/views/search/_sort.html.erb
owenatgov: I'm guessing it was just a design choice 😬 I'll re-work it.

app/views/search/_filters.html.erb
owenatgov: I haven't been able to get this working with the select govpubcomp but I've added a js class to label it so it's clear that it's being used for AA.

app/assets/javascripts/govuk.js
```diff
@@ -15,28 +15,24 @@ $(document).ready(function () {
   var showHide = new ShowHide()
   showHide.init()
 
-  if (document.getElementById('publisher')) {
-    accessibleAutocomplete.enhanceSelectElement({
-      selectElement: document.querySelector('#publisher'),
-      showAllValues: true,
-      preserveNullOptions: true
-    })
+  var autocompleteSelects = document.getElementsByClassName('js-accessible-autocomplete');
```
injms, line 18: From the look of the rest of this file, it looks as if it's been linted with [StandardJS](http://standardjs.com), which is the chosen linter from the [GDS Way rules for JavaScript](https://gds-way.cloudapps.digital/manuals/programming-languages/js.html#linting).

I think the new bits of JavaScript need to be linted as well to ensure that the file is consistent.

```suggestion
  var autocompleteSelects = document.getElementsByClassName('js-accessible-autocomplete')
```

app/views/search/_filters.html.erb
```diff
@@ -1,32 +1,32 @@
 <h2 class="heading-medium"> <%= t('.filter_by') %></h2>
 
-<div class="form-group">
-  <%= label_tag 'publisher', t('.filter_by_publisher'), class: 'form-label' %>
-  <%= select_tag('filters[publisher]', options_for_select(dataset_publishers_for_select, selected_publisher), id: 'publisher', class: 'form-control dgu-filters__filter', include_blank: true) %>
+<div class="govuk-form-group">
+  <%= label_tag 'publisher', t('.filter_by_publisher'), class: 'govuk-label' %>
```
injms, line 4: If DGU uses it, the [GDS Rubocop rules](https://gds-way.cloudapps.digital/manuals/programming-languages/ruby.html#code-formatting) ask for double quotes. 

It can be run to automatically fix some things, so these might be better in another pull request.
```suggestion
  <%= label_tag "publisher", t(".filter_by_publisher"), class: "govuk-label" %>
```

app/views/search/_sort.html.erb
```diff
@@ -1,15 +1,22 @@
...
+      },
+      {
+        text: t('.recent'),
+        value: 'recent',
+        selected: @sort === 'recent'
+      }
+    ]
+  } %>
+  <div class="govuk-visually-hidden">
     <input type="submit" value=<%= t('.accessibility.submit_button') %>/>
```
injms, line 20: This button could be a button component:

```suggestion
    <%= render "govuk_publishing_components/components/button", {
        text: t('.accessibility.submit_button'),
    } %>
```

app/views/search/search.html.erb
```diff
@@ -3,81 +3,74 @@
...
-    </div>
-  </div>
-
-  <div class="grid-row dgu-filters">
-    <div class="column-one-third">
-      <%= render 'filters' %>
-    </div>
+  <main role="main" id="main-content" class="govuk-main-wrapper">
+    <div class="govuk-width-container">
+      <h1 class="govuk-heading-l"> <%= t('.search_results_heading') %></h1>
```
injms, line 8: This could be a [heading component](https://components.publishing.service.gov.uk/component-guide/heading):
```suggestion
      <%= render "govuk_publishing_components/components/heading", {
        text: t('.search_results_heading'),
        heading_level: 1,
        font_size: "l",
      } %>
```

app/views/search/search.html.erb
```diff
@@ -3,81 +3,74 @@
...
+      <div class="govuk-grid-row">
+        <div class="govuk-grid-column-two-thirds dgu-search-box">
```
injms, line 10: Could `dgu-search-box` be removed here as the search component is being used?
```suggestion
        <div class="govuk-grid-column-two-thirds">
```

app/views/search/search.html.erb
```diff
@@ -3,81 +3,74 @@
...
-              <li><%= t('.zero_result_tips.different_words') %></li>
-              <li><%= t('.zero_result_tips.remove_filters') %></li>
-            </ul>
-            <h2 class="heading-medium"> <%= t('.zero_result_tips.older_content') %></h2>
-            <p><%= t('.zero_result_tips.we_can_help') %></p>
-            <p><%= t('.zero_result_tips.expand_your_search') %></p>
+        <div class="govuk-grid-column-two-thirds dgu-results">
+          <%= render 'sort' if @num_results > 1 %>
+          <span class="dgu-results__summary">
+            <span class="bold-small"><%= number_with_delimiter(@num_results) %></span>
```
injms, line 25: We could swap this bespoke CSS out for some styles from the Design System:

```suggestion
            <span class="govuk-body-s govuk-!-font-weight-bold"><%= number_with_delimiter(@num_results) %></span>
```

app/views/search/search.html.erb
```diff
@@ -3,81 +3,74 @@
...
-                <% if dataset.public_updated_at.present? %>
-                  <dd class="last_updated"><%= format_timestamp(dataset.public_updated_at) %></dd>
-                <% else %>
-                  <dd class="last_updated dgu-secondary-text"> <%= t('.meta_data_box.not_applicable') %></dd>
-                <% end %>
-              </dl>
-              <p><%= truncate(strip_tags(dataset.summary), length: 200, separator: ' ') %></p>
+          <% if @num_results.zero? %>
+            <div class="dgu-results__zero">
+              <h2 class="govuk-heading-m"> <%= t('.zero_result_tips.try') %>:</h2>
```
injms, line 31: Another potential heading component:
```suggestion
          <%= render "govuk_publishing_components/components/heading", {
                text: t('.zero_result_tips.try'),
                heading_level: 2,
                font_size: "m",
              } %>
              <h2 class="govuk-heading-m"> <%= t('.zero_result_tips.try') %>:</h2>
```

app/views/search/search.html.erb
```diff
@@ -3,81 +3,74 @@
...
+              <ul class="govuk-list govuk-list--bullet">
+                <li><%= t('.zero_result_tips.different_words') %></li>
+                <li><%= t('.zero_result_tips.remove_filters') %></li>
+              </ul>
+              <h2 class="govuk-heading-m"> <%= t('.zero_result_tips.older_content') %></h2>
```
injms, line 36:
```suggestion
              <%= render "govuk_publishing_components/components/heading", {
                text: t(".zero_result_tips.older_content"),
                heading_level: 2,
                font_size: "m",
              } %>
```

owenatgov:
Rubocop is failing on a few things but I'm going to propose that we bypass this as most of the problems are going to be solved by https://github.com/alphagov/datagovuk_find/pull/790

kentsanggds:
I think the upgrade to Rails 6 and linting fixes have been merged into master so you probably just need to rebase to get tavis to pass

PR #790, opened 2020-07-22
Update rails and other gems

pudinha:
### Before deploying

~~The ElasticSearch connection string on Heroku must be updated to explicitly have the `:443` port, as `gem elasticsearch` started assuming port 9200 by default on version 7.1.
https://github.com/elastic/elasticsearch-ruby/issues/669#issuecomment-503252443~~

not necessary anymore as `gem elasticsearch` version is being kept at 5.x to match the db's major version and maintain compatibility.


config/application.rb
```diff
@@ -28,12 +28,18 @@
...
-    # Settings in config/environments/* take precedence over those specified here.
-    # Application configuration should go into files in config/initializers
-    # -- all .rb files in that directory are automatically loaded.
-
     # Initialize configuration defaults for originally generated Rails version.
-    config.load_defaults 5.1
+    config.load_defaults 6.0
+
+    # Make `form_with` generate non-remote forms.
+    config.action_view.form_with_generates_remote_forms = false
```
pixeltrix, line 35: Does the app use `form_with`? If not doesn't seem much point in having this config.

config/initializers/content_security_policy.rb
```diff
@@ -0,0 +1,30 @@
...
+# If you are using UJS then enable automatic nonce generation
+# Rails.application.config.content_security_policy_nonce_generator = -> request { SecureRandom.base64(16) }
+
+# Set the nonce only to specific directives
+# Rails.application.config.content_security_policy_nonce_directives = %w(script-src)
+
+# Report CSP violations to a specified URI
+# For further information see the following documentation:
+# https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy-Report-Only
+# Rails.application.config.content_security_policy_report_only = true
```
pixeltrix, line 30: The app is using the secure_headers gem so either we need to remove this file or use the policy from the `gov_app_config` gem.

lib/tasks/before_assets_precompile.rake
```diff
@@ -1,3 +1,4 @@
+desc nil
```
pixeltrix, line 1: Was this added automatically or is it a rubocop thing?

config/application.rb
pudinha: this was added by someone else on a previous Rails update (3d421d6). I just moved it here to keep stuff centralized. The app broke when I tried to remove it.

lib/tasks/before_assets_precompile.rake
pudinha: it's just to suppress rubocop. Not pretty, I know...

config/initializers/content_security_policy.rb
pudinha: deleted and rebased on top of master :sparkles:

lib/tasks/before_assets_precompile.rake
edwardkerry: Could we add a little description "Before assets precompile" - redundant but maybe nicer than `nil`?

config/environments/production.rb
```diff
@@ -14,24 +14,24 @@
   config.consider_all_requests_local       = false
   config.action_controller.perform_caching = true
 
-  # Attempt to read encrypted secrets from `config/secrets.yml.enc`.
-  # Requires an encryption key in `ENV["RAILS_MASTER_KEY"]` or
-  # `config/secrets.yml.key`.
-  config.read_encrypted_secrets = true
+  # Ensures that a master key has been made available in either ENV["RAILS_MASTER_KEY"]
+  # or in config/master.key. This key is used to decrypt credentials (and other encrypted files).
+  # config.require_master_key = true
```
edwardkerry, line 20: will we need to implement the new way of managing secrets described here?
https://github.com/rails/rails/pull/30067

pudinha: > If we remove this will we need to implement the new way of managing secrets described here?

This app doesn't use encrypted secrets at all, so no.

But after reading that PR a couple of times, I now see that `/config/secrets.yml` should be deleted, as it's not useful anymore.

PR #794, opened 2020-07-29
Update dataset page markup

owenatgov:
## What
Update the markup, classnames and styling on the individual dataset view and associated components and partials. Whilst there are mostly intended to be very few discernible difference between this PR and production, there are a number of visual changes introduced as a consequence of updating parts of the frontend tooling:

- The search box to the right of the metadata component is using the govuk search component with a blue button instead of a black one
- Several instances of data lists have been wrapped in the govuk details component instead of using custom styling to introduce the grey left-hand border synonymous with govuk detail components, which changes the width of some of these sections
- The "show more" link for if the datafile table is too long has been replaced by a secondary govuk button

## Why

**Card:** https://trello.com/c/h0XbWxq8/233-update-markup-on-find-dataset-page

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -16,6 +16,7 @@
 //= require govuk_toolkit
 //= require accessible-autocomplete/dist/accessible-autocomplete.min
 //= require govuk_publishing_components/dependencies
+//= require govuk_publishing_components/modules
 //= require govuk_publishing_components/all_components
 //= require_tree ./organograms/lib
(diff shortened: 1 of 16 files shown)


kr8n3r:
the tendency use to be to use `govuk-spacing(n)` for margin and padding. not sure if that's now been changed

owenatgov:
I'm closing this PR as it's very large and encompasses a large number of features that've been stuffed into this one view. Will break down into a number of smaller PRs for easier digestion.

PR #796, opened 2020-08-04
Update markup for data preview view

owenatgov:
## What
Update the markup, classnames and styling on the datafile preview view. Whilst there are mostly intended to be no discernible difference between this PR and production, it does replace the custom preview table with the govuk table component.

### Before
Screenshot 2020-08-04 at 10 29 10

### After
Screenshot 2020-08-04 at 10 28 36

## Why

**Card:** https://trello.com/c/jGqwrZ3n/248-update-markup-on-find-preview-pages

### DIFF
--- a/app/helpers/previews_helper.rb
+++ b/app/helpers/previews_helper.rb
@@ -6,4 +6,8 @@ module PreviewsHelper
       t(".download_file")
     end
   end
+
+  def is_numeric(string)
+    string.match?(/\d/)
+  end
 end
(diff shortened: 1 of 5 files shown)


owenatgov:
> Looks good. Just one suggestion, if you think the majority of table cells will contain numeric data it might be worth setting that as a default. It's a tough call because it's essentially anonymous data that could be anything. If you set it as numeric then it will line things up by right aligning them, but then that will look odd for non-numeric fields, but likewise assuming non-numbers makes numeric comparison harder at a glance.

That's a good shout. I'm gonna see if there's a way I can check this on the frontend.

app/helpers/previews_helper.rb
```diff
@@ -6,4 +6,8 @@ def download_text(dataset, datafile)
       t(".download_file")
     end
   end
+
+  def is_numeric(string)
+    true if Float(string) rescue false
```
maxgds, line 11: Nice solution. It's still going to fail on things like "10k" or "£50,000" but it's probably better than blind assumption.

kentsanggds:
I think the upgrade to Rails 6 and linting fixes have been merged into master so you probably just need to rebase to get tavis to pass

PR #800, opened 2020-08-10
Correct Rubocop errors

edwardkerry:
Corrects a number of Rubocop errors, see commits for individual fixes


edwardkerry:
Handled in https://github.com/alphagov/datagovuk_find/pull/790

PR #808, opened 2020-08-25
Add CKAN maintenance page

kentsanggds:
## What

Add a maintenance page for CKAN downtime when we do the upgrade work, response status code will be 503

image

## Reference 

https://trello.com/c/Mz0Psh36/264-create-503-page-on-dgufind


thomasleese (approved):

risicle:
What url will this be published at?

PR #832, opened 2020-10-06
Update dataset page markup for foundational features

owenatgov:
## What
Update the markup, classnames and styling on the individual dataset view index as well as the breadcrumb and metadata partials.

## Why

This PR was created as part of an effort to [break down the existing dataset view PR](https://github.com/alphagov/datagovuk_find/pull/794) which had grown too large to safely review in one go as it contains a large number of individual features.

## Visual changes
The only visual change this introduces is that the search box to the right of the metadata component is using the govuk search component with a blue button instead of a black one.

**Card:**  https://trello.com/c/h0XbWxq8/233-update-markup-on-find-dataset-page


app/assets/stylesheets/components.scss
```diff
@@ -7,7 +7,7 @@
 
 .dgu-highlight {
   background: $grey-3;
```
maxgds, line 9: Probably worth updating your colours while you're in here, eg govuk-colour("light-grey")

app/assets/stylesheets/components.scss
```diff
@@ -7,7 +7,7 @@
...
-  padding: em(2) em(5) em(2) em(5);
+  padding: 2px govuk-spacing(1);
```
maxgds, line 10: What's the need for the 2px? Is it needed anywhere else? Ideally we don't have bare numbers in the main css, and might assign it to a var, but maybe not worth it for one outlier.

app/views/datasets/_breadcrumb.html.erb
```diff
@@ -1,26 +1,25 @@
...
-              <% else %>
-                <%= link_to t('.search'), "/search?#{@referrer}", class: 'govuk-link' %>
-              <% end %>
-            </li>
-            <li aria-current="page">
-              <%= shorten_title(@dataset.title) %>
-            </li>
-          </ol>
-        </nav>
+  <section id="dgu-breadcrumb">
```
maxgds, line 2: We tend to use BEM conventions which avoids using ID selectors. Is there a reason why you've switched to or introduced IDs in a few places?

maxgds (commented):
Some suggestions/comments added, nothing blocking. I haven't run the code. 

One last question - have you checked the before/after at tablet and mobile view sizes? 

If you want to check out the comments, and maybe later pair to run me through it working I'd be happy to approve this as it's definite steps forward.

owenatgov:
Closing this for now to unclog the DGU PR list a bit and as I don't know when I'll be able to get to this

PR #842, opened 2020-10-21
Add auth file to prove ownership of data.gov.uk

JonathanHallam:
We need to ask google to purge their cached copies of some files. In order for them to trust us we need to prove ownership of data.gov.uk. This adds a file that can be scraped by google to prove we are who we say we are.


gclssvglx (approved):
LGTM

PR #847, opened 2020-11-03
Restyle table to be clearly separated on page and focusable for keyboard users

owenatgov:
## What
Restyles the table element on the data preview view to wrap it in a set of parent elements that make it more clearly visually distinct and add `tabindex="0"`, allowing the table to be focused when tabbing through the page and for overflowing tables to be navigable via the left and right keys.

## Why
Besides the quality of life change for sighted users, ensuring that tables are navigable for keyboard users is necessary for WCAG 2.1 SC 2.1.1. This PR is also to make completely sure that our preview view doesn't fail SC 1.4.10.

### Before
Screenshot 2020-11-03 at 18 05 39

### After
Screenshot 2020-11-03 at 15 40 09
Screenshot 2020-11-03 at 15 40 19

### DIFF
--- a/app/views/previews/show.html.erb
+++ b/app/views/previews/show.html.erb
@@ -62,10 +62,12 @@
       <div class="govuk-grid-column-full">
         <section id="dgu-preview">
           <div class="dgu-datafile-preview">
-            <%= render "govuk_publishing_components/components/table", {
-              head: table_headings,
-              rows: table_rows
-            } %>
+            <div class="dgu-datafile-preview__inner" tabindex="0">
+              <%= render "govuk_publishing_components/components/table", {
+                head: table_headings,
+                rows: table_rows,
+              } %>
+            </div>
           </div>
         </section>
 
--- a/app/views/previews/show.html.erb
+++ b/app/views/previews/show.html.erb
@@ -72,7 +72,7 @@
         </section>
 
         <% if @dataset.organogram? %>
-          <section id="organogram" data-csv-url="<%= @datafile.url %>">
+          <section id="organogram" class="dgu-datafile-preview__organogram" data-csv-url="<%= @datafile.url %>" tabindex="0">
           </section>
         <% end %>
       </div>
--- a/app/assets/stylesheets/components.scss
+++ b/app/assets/stylesheets/components.scss
@@ -269,7 +269,19 @@
 // preview page
 
 .dgu-datafile-preview {
-  overflow: auto;
+  border: 5px solid govuk-colour("light-grey");
+  margin-bottom: govuk-spacing(6);
+
+  &__inner {
+    border: 1px solid govuk-colour("mid-grey");
+    padding: govuk-spacing(2) govuk-spacing(4);
+    overflow: auto;
+
+    &:focus {
+      border: 3px solid govuk-colour("black");
+      outline: 3px solid govuk-colour("yellow");
+    }
+  }
 }
 
 // =============================================================================
--- a/app/assets/stylesheets/components.scss
+++ b/app/assets/stylesheets/components.scss
@@ -282,6 +282,11 @@
       outline: 3px solid govuk-colour("yellow");
     }
   }
+
+  &__organogram:focus {
+    border: 3px solid govuk-colour("black");
+    outline: 3px solid govuk-colour("yellow");
+  }
 }
 
 // =============================================================================


app/views/previews/show.html.erb
```diff
@@ -62,10 +62,12 @@
       <div class="govuk-grid-column-full">
         <section id="dgu-preview">
           <div class="dgu-datafile-preview">
-            <%= render "govuk_publishing_components/components/table", {
-              head: table_headings,
-              rows: table_rows
-            } %>
+            <div class="dgu-datafile-preview__inner" tabindex="0">
```
chris-gds, line 65: I ran into some issues testing locally, semi-workaround using [this page](https://data.gov.uk/dataset/2add7628-ba06-480e-8c38-8f6da502743c/spend-over-25-000-in-nhs-leeds/datafile/3bbdc494-ecef-4b54-bb01-a1f099f86d00/preview).

So on the surface this seems OK but on testing: VoiceOver now reads out all tabular content on initial focus, I wouldn’t think this is desirable and doesn't occur without this..🤔

I’m wondering in terms of 2.1.1 if this change is needed in the sense that valid tabular markup is keyboard accessible by default. The main concern is the correct markup + data-population use, (eg might a caption be needed? Might the first column be classed as a header?)

Also, from a [reflow (SC 1.4.10.) POV “without requiring scrolling in two dimensions”](https://www.w3.org/WAI/WCAG21/Understanding/reflow.html) the table isn’t responsive so this fails, the user has to scroll horizontally to view this on mobile / zoomed.

There’s also visible [divergence from the Design system](https://design-system.service.gov.uk/components/table/), eg the primary GDS font seems to be missing within this app?

chris-gds: Visual VO representation upon focus:

Screenshot 2020-11-04 at 09 23 55

injms: > Also, from a [reflow (SC 1.4.10.) POV “without requiring scrolling in two dimensions”](https://www.w3.org/WAI/WCAG21/Understanding/reflow.html) the table isn’t responsive so this fails, the user has to scroll horizontally to view this on mobile / zoomed.

I've not seen a perfect or, to be honest, even a good responsive table that doesn't involve breaking the table's internal relationships - such as a cell that has a row and a column heading. This makes it harder for screenreaders etc to navigate through the table. 

Pragmatically, I think that this is a good enough improvement to be merged in as it does make the page more accessible.

> Side note: noticed, there’s also visible [divergence from the Design system](https://design-system.service.gov.uk/components/table/), eg the primary GDS font seems to be missing within this app.

Transport (the custom font used on GOV.UK, the services, and the blogs) isn't used on DGU.

owenatgov: >VoiceOver now reads out all tabular content on initial focus, I wouldn’t think this is desirable and doesn't occur without this..🤔

That's less than ideal 😬  I'll experiment with blocking that

>I’m wondering in terms of 2.1.1 if this change is needed in the sense that valid tabular markup is keyboard accessible by default. The main concern is the correct markup + data-population use, (eg might a caption be needed? Might the first column be classed as a header?)

I disagree. Before, there was no way for keyboard only users to actually view tables without a mouse click. There's contention around if this is still a failure or not but I'd rather be safe than sorry. I agree that the real fix should be the data, but this is going to require us to very closely manage the data that publishers publish, which we don't have the resource to do right now.

>Also, from a reflow (SC 1.4.10.) POV “without requiring scrolling in two dimensions” the table isn’t responsive so this fails, the user has to scroll horizontally to view this on mobile / zoomed.

There's a note on that success criterion:

```
Examples of content which require two-dimensional layout are images, maps, diagrams, video, games, presentations, data tables, and interfaces where it is necessary to keep toolbars in view while manipulating content.
```

This technically makes this preview exempt as it counts as a data table. I ran this past the accessibility channel and they said that whilst technically we can argue it away, some usability improvements will give us better chances to avoid getting pulled up on it. Again, ideally the data wouldn't be so gross and unpredictable (or this view would be retired) but we can't get around that right now.

>Side note: noticed, there’s also visible divergence from the Design system, eg the primary GDS font seems to be missing within this app.

This is intentional. There's some annoying politics around if data.gov.uk counts as a gov.uk site. [Currently, the data.gov.uk URI isn't allowed to be used under the legislation for the GDS font](https://designnotes.blog.gov.uk/2015/03/11/can-i-use-the-gov-uk-fonts/).

chris-gds: > I've not seen a perfect or, to be honest, even a good responsive table that doesn't involve breaking the table's internal relationships - such as a cell that has a row and a column heading. This makes it harder for screenreaders etc to navigate through the table.

[Responsive tables](https://adrianroselli.com/2017/11/a-responsive-accessible-table.html) aren't straight forward f'sure (venturing into epic world...), pragmatically and in context on this project makes sense.

>That's less than ideal 😬 I'll experiment with blocking that

Yeah, didn't _sound_ ideal but could be a simple way to avoid might just be for VO.

>I disagree. Before, there was no way for keyboard only users to actually view tables without a mouse click. 

Maybe I'm missing a trick here, a user can navigate through the doc not just via focusable elements? 🤔  

> There's contention around if this is still a failure or not but I'd rather be safe than sorry. I agree that the real fix should be the data, but this is going to require us to very closely manage the data that publishers publish, which we don't have the resource to do right now.

*nods* 👍 


>This technically makes this preview exempt as it counts as a data table. I ran this past the accessibility channel and they said that whilst technically we can argue it away, some usability improvements will give us better chances to avoid getting pulled up on it. Again, ideally the data wouldn't be so gross and unpredictable (or this view would be retired) but we can't get around that right now.

Gotcha - this unpredictable data aspect makes this hard to factor in.

> Transport (the custom font used on GOV.UK, the services, and the blogs) isn't used on DGU.
&
> This is intentional. There's some annoying politics around if data.gov.uk counts as a gov.uk site

Ah.

owenatgov: >Maybe I'm missing a trick here, a user can navigate through the doc not just via focusable elements? 🤔

So, if you go [here](https://data.gov.uk/dataset/a4bcf3a1-8be1-46c8-aa01-b009841d4829/organogram-of-staff-roles-salaries/datafile/b6ed9853-25d7-470f-aa10-0c825bf1e334/preview) and navigate the page by keyboard only using tabbing and arrow keys, there's no way to see the data that's hidden behind a scroll bar in the data table without explicitly clicking into that table sub-view.

This is putting aside the really grubbing design of that page where it's not clear that the scrollbar under the table is associated with the table because of poor space management.

chris-gds: Riiigght - gotcha, course! I was coming at this from from a VO angle on the other page `CAPs` + `>` Yup that's an issue 😁 👍🏼

chris-gds (approved):

PR #854, opened 2020-11-18
Update elasticsearch 7

kentsanggds:
## What

As elasticsearch is being upgraded the mapping has changed so that has been updated to work with it and the way it tracks the total number of results has changed which means that we have to pass in an additional arg for search queries - `track_total_hits`

Also in this PR we are changing the space that Staging find is deployed to.

## Reference 

https://trello.com/c/xjutCWHA/2215-upgrade-elasticsearch-version-on-paas


thomasleese (approved):
I don't fully understand the differences in this version of Elasticsearch, but it looks sensible and the tests pass.

PR #857, opened 2020-11-23
Relax WebMock elasticsearch host constraint

pixeltrix:
There are a range of elasticsearch host names in use by so allow any host that starts with elasticsearch but don't allow FQDN values such as elasticsearch.hosting.tld.


kevindew (approved):

benthorner (approved):
🚀

PR #860, opened 2020-11-26
Use full uri to WebMock allow option

pixeltrix:
Whilst strings are only matched against the host, regexes are matched against the full uri so we need to add protocol, port and only anchor the uri at the beginning.

### DIFF
--- a/spec/rails_helper.rb
+++ b/spec/rails_helper.rb
@@ -33,7 +33,10 @@ Dir[Rails.root.join("spec/support/**/*.rb")].sort.each { |f| require f }
 
 FactoryBot.find_definitions
 
-WebMock.disable_net_connect!(allow_localhost: true, allow: /^elasticsearch[-a-z0-9]*$/)
+WebMock.disable_net_connect!(
+  allow_localhost: true,
+  allow: Rails.configuration.elasticsearch["host"],
+)
 
 RSpec.configure do |config|
   # RSpec Rails can automatically mix in different behaviours to your tests


spec/rails_helper.rb
```diff
@@ -33,7 +33,7 @@
 
 FactoryBot.find_definitions
 
-WebMock.disable_net_connect!(allow_localhost: true, allow: /^elasticsearch[-a-z0-9]*$/)
+WebMock.disable_net_connect!(allow_localhost: true, allow: %r{^https?://elasticsearch[-a-z0-9]*(?::\d+)?/})
```
benthorner, line 36:
```suggestion
WebMock.disable_net_connect!(allow_localhost: true, allow: %r{.*elasticsearch.*})
```

Minor: something simpler should be suitable for testing, noting it's unclear that we'll be making other, stubbed calls to anything with "elasticsearch" in the URL.

benthorner (approved):
Looks good to me - I assume you've tested it out.

spec/rails_helper.rb
pixeltrix: My only concern with using a general regex like that is it would allow running tests against a live database if it happened to have elasticsearch in the url, however that risk is relatively low so happy to change it if you want.

pixeltrix:
@benthorner I've updated this to match the style in DGU Publish which reads the config from elasticsearch.yml - this seems a much cleaner solution than messing about with regexes

benthorner (approved):
Good idea - much clearer 👍.

PR #885, opened 2021-01-28
Update README.md

owenatgov:
## What
Update the readme to note alternative means to run the repo locally, specifically highlighting docker-ckan which is the preferred means.

## Why
Prompted by needs to [ensure that all our documentation is up to date](https://github.com/alphagov/govuk-ithc-documentation/issues/111).

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,7 +5,16 @@
 
 This repository contains the beta-stage frontend component of data.gov.uk
 
-## Prerequisites
+## How to run this repo locally
+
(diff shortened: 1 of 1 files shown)


erino:
@owenatgov thanks for this, can you also check that this file is up to date https://github.com/alphagov/govuk-ithc-documentation/blob/master/datagovuk_find.md

kentsanggds (approved):

PR #903, opened 2021-02-26
Add govuk grid classes to header and nav items

owenatgov:
## What
Adds govuk-grid classes to the header.

## Why
This fixes a layout issue within the header. Find is the only app consuming the separately exposed `navigation_items` sub-component within the [layout header component](https://components.publishing.service.gov.uk/component-guide/layout_header) because it can't legally use the crown logo. Therefore it's missed out on some markup updates to the layout header.

Merging this change will mean that DGU is finally ready for a new release, including using the new, ICO compliant design system cookie banner, minimising risk of legal action against us via a blind spot like DGU.

This bug isn't visible on live but you can see it on staging [here](https://find-data-beta-staging.cloudapps.digital/).

### Before
Screenshot 2021-02-26 at 11 29 35

### After
Screenshot 2021-02-26 at 11 34 31

### DIFF
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -31,12 +31,16 @@
 
     <header role="banner" class="govuk-header" data-module="govuk-header">
       <div class="govuk-header__container govuk-width-container">
-        <div class="govuk-header__logo">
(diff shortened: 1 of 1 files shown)


bilbof (approved):
LGTM! 🎉

PR #971, opened 2021-07-27
Fix Style/RedundantRegexpEscape linting offence

kevindew:
This is auto-corrected by Rubocop. The dash character does not need
escaping.

This is to fix the main branch after I snookered myself by merging in
a PR before rebasing it against the main branch with updated linting
rules.


1pretz1 (approved):
Good spot Kev! 🟢

PR #973, opened 2021-08-03
Add integration deployment files

kentsanggds:
## What

Add files for deployment to integration stack on PaaS.


deborahchua (approved):

brucebolt (approved):
It might be an idea to move the manifests into a separate directory at some point, so they don't clutter the root?

PR #976, opened 2021-08-06
Fix organogram CSP error

richardTowers:
Some organograms are hosted on ckan.publishing.service.gov.uk, instead
of s3-eu-west-1.amazonaws.com.

The previews for these organograms are currently broken because the
content security policy prevents the JavaScript from dowloading the
files from the CKAN domain.

I confirmed that the CSP was the only issue by disabling CSP in my
brower (using a plugin) and confirming that the broken previews worked
correctly.

Since the CSP already permits all of S3 eu-west-1 in the connect_src,
adding CKAN to the CSP feels like a very small piece of extra security
attack surface. And it should be a quick way to fix the bug where some
organogram previews don't show up.

I tried to add a test for this, but rails controller tests don't execute
enough of the stack for the SecureHeaders gem to do its thing and set
the CSP header. It might be possible to test with a feature test, but
that feels like overkill.

I also renamed `CKAN_REDIRECTION_URL` to `CKAN_DOMAIN` to
make it clear how it's being used (the CSP needs it to be a domain, not
a full URL, which fortunately it already is).

### DIFF
--- a/config/environments/development.rb
+++ b/config/environments/development.rb
@@ -66,6 +66,4 @@ Rails.application.configure do
   # config.action_cable.disable_request_forgery_protection = true
 
   config.zendesk = nil
-
-  config.ckan_redirection_url = ENV["CKAN_REDIRECTION_URL"]
 end
--- a/app/controllers/previews_controller.rb
+++ b/app/controllers/previews_controller.rb
@@ -1,7 +1,7 @@
 class PreviewsController < ApplicationController
   def show
     append_content_security_policy_directives(
-      connect_src: %w[s3-eu-west-1.amazonaws.com],
+      connect_src: ["s3-eu-west-1.amazonaws.com", ENV["CKAN_DOMAIN"]].compact,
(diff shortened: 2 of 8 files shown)


config/environments/development.rb
```diff
@@ -66,6 +66,4 @@
   # config.action_cable.disable_request_forgery_protection = true
 
   config.zendesk = nil
-
-  config.ckan_redirection_url = ENV["CKAN_REDIRECTION_URL"]
```
ChrisBAshton, line 70: Nitpick: ideally this (and production.rb) would be a separate commit, as everything else is just a rename.

ChrisBAshton (approved):
LGTM 👍 nice one!

config/environments/development.rb
richardTowers: Yeah, fair cop. I'm not going to split this commit, but I'll try to be more granular next time.

PR #1053, opened 2022-03-17
Migrate elasticsearch opensearch

kentsanggds:
## What

Migrate from elasticsearch to opensearch as elasticsearch is being withdrawn from PaaS


whpearson (approved):
Is the same as other application

PR #1112, opened 2022-10-18
Remove dot (.) prefix from formats

kentsanggds:
## What

Some resources were set with the . prefix, e.g. .csv, this would cause csv prefix links to not show. 
So removing the `.` from the format will enable previews again, this has been tested on a local dev stack


KludgeKML (approved):
LGTM

PR #1169, opened 2023-05-19
Update ruby version to 3.2.0

kentsanggds:
Update to ruby 3.2

- Had to monkeypatch kaminari due to an issue with how it handles `to` count.


spec/rails_helper.rb
```diff
@@ -71,3 +71,30 @@
...
+
+# Setup chrome headless driver
+Capybara.server = :puma, { Silent: true }
+
+Capybara.register_driver :chrome_headless do |app|
+  options = ::Selenium::WebDriver::Chrome::Options.new
+
+  options.add_argument("--headless")
+  options.add_argument("--no-sandbox")
+  options.add_argument("--disable-dev-shm-usage")
```
sengi, line 83: Good call! 👍

spec/rails_helper.rb
```diff
@@ -71,3 +71,30 @@
...
 end
+
+# Setup chrome headless driver
+Capybara.server = :puma, { Silent: true }
+
+Capybara.register_driver :chrome_headless do |app|
+  options = ::Selenium::WebDriver::Chrome::Options.new
+
+  options.add_argument("--headless")
+  options.add_argument("--no-sandbox")
```
sengi, line 82: Don't worry about it for this PR, but just fyi it should be possible to run in the sandbox. Might need to switch to our govuk-ruby-images though (hence definitely not worth holding up this PR).

Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM ruby:2.7.6
+FROM ruby:3.2.0
```
sengi, line 1: Any reason not to use 3.2.2 or 3.2 (i.e. latest security patch)?

sengi (approved):

Dockerfile
kentsanggds: Not sure, just that I've tested it against ruby 3.2.0 and have sorted out the dependencies so don't want to mess around with it any more

sengi: No probs 👍 Should be a drop-in replacement so hopefully a very simple update when you have a spare moment!

PR #1173, opened 2023-08-03
Update `govuk_test` to v3.0.1

robinjam:
This removes the dependency on the `webdrivers` gem, which was causing test failures on CI (and in Docker).

Also updates the Chrome WebDriver to the latest version (the latest version of Chrome does not yet have a corresponding WebDriver, so we have to hardcode a slightly older point release)


kentsanggds (approved):
LGTM

PR #1172, opened 2023-08-03
Disallow `/search` in `robots.txt`

robinjam:
[Trello card](https://trello.com/c/9Y1k5t6J/3271-noindex-datagovuk-search-result-pages-2)

Removing the trailing slash, as our search URLs (e.g. https://www.data.gov.uk/search?q=test+registration) do not include this slash.

PR #1174, opened 2023-08-07
Mark search results pages as `noindex`

robinjam:
[Trello card](https://trello.com/c/9Y1k5t6J/3271-noindex-datagovuk-search-result-pages-2)

Supersedes #1172 

We set `noindex` to prevent search engines from indexing the search results pages themselves, but we do not set `nofollow`, as we want search engines to be able to use the search results pages to discover other pages on DGU. This is important because DGU doesn't have a sitemap.


kentsanggds (approved):
LGTM

PR #1176, opened 2023-08-07
Add Google site verification for www.data.gov.uk

robinjam:
data.gov.uk is already covered, but Google sees it as a separate domain


kentsanggds (approved):
LGTM

PR #1195, opened 2023-11-27
Update build image tag to run against v prefix

kentsanggds:
This will trigger deployments to the kubernetes cluster when tagged releases are prefix with v

## Reference

https://trello.com/c/2qXYzZYb/1239-switchover-paas-apps-to-k8s-cluster


nimalank7 (approved):
LGTM!

PR #1198, opened 2023-11-30
Use govuk-ruby-images and disable pidfile.

sengi:
- Stop Rails/Puma from writing pidfiles, which was causing problems in production.
- Remove unneeded tmp_restart Puma plugin.
- Remove disused config files left over from when we ran Find on CloudFoundry.
- Refactor Dockerfile to use [govuk-ruby-images](https://github.com/alphagov/govuk-ruby-images). This gives us a bunch of stuff for free, such as:
    - jemalloc (more consistent memory usage, which helps in production)
    - common packages preinstalled so less boilerplate
    - two-stage build so output image is smaller with no leftover cruft from the build

Tested: builds, runs and serves at least the homepage ok:

```sh
ARCH=$(arch) DOCKER_TAG=find APP=find GITHUB_SHA=main docker/build-image.sh; 
docker run -it --rm -e ES_HOST=x -e SECRET_KEY_BASE=x \
  -e RAILS_SERVE_STATIC_FILES=1 -p 3000:3000 \
  ghcr.io/alphagov/find:main
```

Was also able to build and run with the [k3s setup in dgu-charts](https://github.com/alphagov/govuk-dgu-charts/?tab=readme-ov-file#quickstart-for-local-development-cluster).


sengi:
@kentsanggds I think 736ebb6 might fix the trouble you were having trying this out.

samsimpson1 (approved):

PR #1205, opened 2024-01-16
Fix create-pr-on-tags.yaml

kentsanggds:
Previous code to get the github ref is no longer working so use `git describe`

### DIFF
--- a/.github/workflows/create-pr-on-tags.yaml
+++ b/.github/workflows/create-pr-on-tags.yaml
@@ -14,7 +14,7 @@ jobs:
     steps:
       - uses: actions/checkout@v4
       - name: Set GH_REF
-        run: echo "GH_REF=`echo $(git fetch -t -q | sort --version-sort | tail -n1)`" >> $GITHUB_ENV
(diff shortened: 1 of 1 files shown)


.github/workflows/create-pr-on-tags.yaml
```diff
@@ -14,7 +14,7 @@ jobs:
     steps:
       - uses: actions/checkout@v4
       - name: Set GH_REF
-        run: echo "GH_REF=`echo $(git fetch -t -q | sort --version-sort | tail -n1)`" >> $GITHUB_ENV
+        run: echo "GH_REF=`echo $(git fetch -t -q && git describe --tags)`" >> $GITHUB_ENV
```
sengi, line 17: Does this always do what you want when there are multiple tags at head? (Is that something that can happen here?)

sengi (approved):

.github/workflows/create-pr-on-tags.yaml
kentsanggds: The same code is being used for datagovuk_publish, and that appears to be working well

kentsanggds: ok, I've found a better one which I've just tested on my machine

PR #1217, opened 2024-02-15
Heed GA_TRACKING_ID at runtime.

sengi:
The Google Analytics tracking code - which is supposed to be configurable via an environment variable - was being burned into a Rails asset file at build-time.

Fix it so that the environment variable actually works as expected.

Fixes https://github.com/alphagov/govuk-dgu-charts/issues/128.


samsimpson1 (approved):

sengi:
Ah I'd missed a Sprockets preprocessor magic comment that was referring to the deleted ERB template. Deary me.

PR #1237, opened 2024-03-07
Add dev.Dockerfile to enable running on local dev cluster

kentsanggds:
dev.Dockerfile extends the production Dockerfile to include development environment dependencies that would be useful to use in a local cluster to aid running tests and debugging issues.


sengi (approved):
🚀

PR #1250, opened 2024-04-04
Point user to an alternative way to get urgent support

AgaDufrat:
The support form was removed but we need a way for legitimate users (eg those asking to become data publishers) to contact us.


nicholsj (approved):
LGTM (although I can't preview the page)

PR #1251, opened 2024-04-08
Add redirect of support/new to /support

baisa:
To stop bots from being able to raise spam tickets.

[Trello card] (https://trello.com/c/TrY9IEND/2300-dgu-form-spam)

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -30,6 +30,7 @@ Rails.application.routes.draw do
     get "publishers"
     get "site-changes", to: :site_changes
     get "support"
+    get "support/new", to: redirect("/support")
     get "terms"
     get "ckan_maintenance"
   end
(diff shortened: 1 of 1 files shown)


config/routes.rb
```diff
@@ -39,7 +40,6 @@
 
   get "search/", to: "search#search"
 
-  resources :tickets, only: %i[new create]
   get "tickets/confirmation", to: "tickets#confirmation"
```
ChrisBAshton, line 43: I'd also remove this route - and add a redirect route for `https://www.data.gov.uk/tickets/new` -> `https://www.data.gov.uk/support` 🙏 thank you!

baisa: Does this look all right?

config/routes.rb
```diff
@@ -39,7 +40,8 @@
 
   get "search/", to: "search#search"
 
-  resources :tickets, only: %i[new create]
+  get "tickets/new", to: redirect("https://www.data.gov.uk/support")
```
ChrisBAshton, line 43:
```suggestion
  get "tickets/new", to: redirect("/support")
```

I think that's correct 🙏

config/routes.rb
ChrisBAshton: (Line 43 still needs deleting)

ChrisBAshton (approved):
LGTM 👍 please check the redirects work on integration before promoting to staging/production (as per https://docs.publishing.service.gov.uk/manual/data-gov-uk-deployment.html) - thanks!

PR #1256, opened 2024-04-19
Implement Google Tag Manager

PeterHattyar:
This adds GA4 tracking to DGU Find. It is configured with the below environment variables:
- GOOGLE_TAG_MANAGER_ID
- GOOGLE_TAG_MANAGER_AUTH (integration only)
- GOOGLE_TAG_MANAGER_PREVIEW (integration only)

Values are defined in https://github.com/alphagov/govuk-dgu-charts/pull/179

Trello card: https://trello.com/c/VdzhWk09/3456-migrate-datagovuk-page-views-tracking-from-ua-to-ga4-3

PR #1259, opened 2024-04-30
Setting up Google Tag Manager in Content Security Policy

PeterHattyar:
Setting up Google Tag Manager in Content Security Policy, so that the script is allowed to be run.

As it can see below, this is currently being prevented by content policy in Find, as it is in its Rails default state:
GTM
It is visible that Google Tag Manager is present
Issue
But the script is prevented to run by default policy.

Related to PR: https://github.com/alphagov/datagovuk_find/pull/1256

Trello card: https://trello.com/c/VdzhWk09/3456-migrate-datagovuk-page-views-tracking-from-ua-to-ga4-3


MahmudH (approved):
Thanks for adding the screenshots. LGTM!

PR #1277, opened 2024-06-11
Disable session store in order to allow fastly to cache

kentsanggds:
Also CSRF no longer needed as support forms have been disabled.


MahmudH (approved):
👍🏽

PR #1313, opened 2024-09-27
Add Solr and basic search page

deborahchua:
This PR contains the initial set up for using Solr and performs a basic search to display the first page of results to the user when the page is loaded. It can be reached at `/search/solr` to allow OpenSearch to still be the default search method.


kentsanggds (approved):
LGTM

PR #1314, opened 2024-10-04
Add sort filters to search page

deborahchua:
By default the results returned are "best match" and the sort option in the filter is automatically selected.

When the results are sorted by "most recent" we want to sort using the `public_updated_at` (the equivalent field in Solr is `metadata_modified`) and ordered by desc. This is passed as the `sort` parameter in the Solr query.

This has been tested in Integration.

Sorted by "Best match":
Screenshot 2024-10-07 at 12 17 58

Sorted by "Most recent":
Screenshot 2024-10-07 at 12 18 09

### DIFF
--- a/spec/features/solr_search_page_spec.rb
+++ b/spec/features/solr_search_page_spec.rb
@@ -3,43 +3,66 @@ require "rails_helper"
 RSpec.describe "Solr Search page", type: :feature do
   let(:results) { File.read(Rails.root.join("spec/fixtures/solr_response.json").to_s) }
 
-  before do
-    allow(Search::Solr).to receive(:search).and_return(JSON.parse(results))
+  describe "Default search page behaviour" do
+    before do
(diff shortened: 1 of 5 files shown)


spec/features/solr_search_page_spec.rb
```diff
@@ -3,43 +3,66 @@
...
+  describe "Default search page behaviour" do
+    before do
+      allow(Search::Solr).to receive(:search).and_return(JSON.parse(results))
+    end
 
-  scenario "Displays search heading" do
-    visit "/search/solr"
-    expect(page).to have_css("h1", text: "Search results")
-  end
+    scenario "Displays search heading" do
```
kentsanggds, line 11: Is this indent intentional?

deborahchua: Yes I've grouped these tests into a separate `describe` block to keep them separate from the new tests added.

kentsanggds (approved):
LGTM

Issue #1335, opened 2024-10-28
Topic filter should only include topics for datasets that are included in the search results


deborahchua:
If a user has entered a search term and/or has selected some filters, we should expect the `topics` filter to only include topics of datasets that are included in the search results.

The closest thing we have available as the topic name is the `extras_theme-primary` field. We can use it as a facet in the Solr query, and only return any where the minimum count is 1. Unfortunately it does not return the names we want to display in the UI, for example:

```
[
  "environ",
  "govern",
  "governmentspend",
  "spend",
  "and",
  "busi",
  "businessandeconomi",
  "economi",
  "health"
]
```

The list also includes duplication of names that are split by whitespace e.g. the topic `Business and economy` is returned as `busi`, `and`, `economi` as well as `businessandeconomi`. We need to be able to map each topic to it's
display name.

Searching for "Animal health" with no filters selected:
![Screenshot 2024-10-28 at 14 51 27](https://github.com/user-attachments/assets/e183b6c3-5331-4d13-b502-8ebc26013d03)

We could created another custom index (similar to `site_id:dgu_organisations`, see https://github.com/alphagov/ckanext-datagovuk/pull/1852) and the response included from Solr would look something like:

```
docs:[
        { name: "govern", title: "Government" },
        { name: "governmentspend", title: "Government spending" },
        { name: "businessandeconomi", title: "Business and economy" },
      ]
```
We can then map each name to it's appropriate display name.
The topics are hardcoded on the homepage and don't change so we would only need to index the information once.

This query lists all available topics and a count for each in the Solr UI - `/solr/#/ckan/query?q=*:*&q.op=OR&indent=true&facet=true&facet.sort=count&facet.mincount=1&fl=name&rows=0&facet.field=extras_theme-primary`

![Screenshot 2024-10-28 at 15 27 12](https://github.com/user-attachments/assets/279c243e-1346-463c-ac8b-2c540c871678)

See WIP - https://github.com/alphagov/datagovuk_find/commit/23871c2319fa98dff230cd11d991d2d90cb77937 on branch`topic-format-filters`

For more information on faceting see https://solr.apache.org/guide/8_11/faceting.html

deborahchua:
We decided we should hardcode the topics and map them to their corresponding name/display name, as there are only 10-12 and don't change.

AgaDufrat:
[...]

PR #1344, opened 2024-11-06
Update Solr topic filter, fix removing filters and persist keywords

AgaDufrat:
### Changes

- Use Presenter, correct design patter which simplifies further dev work, instead of Helper.
- Refactors and configuration changes to improve performance (tests were failing due to timeouts)
- **Topic filter only includes topics of returned datasets**

Before  | After
------------- | -------------
 Screenshot 2024-11-07 at 12 55 25 | Screenshot 2024-11-07 at 12 38 10

- **Fix removing filters**
<kbd>removing_filters</kbd>

- **Don’t clear search keywords when filters are applied**
<kbd>apply_filters
</kbd>

Closes https://github.com/alphagov/datagovuk_find/issues/1335


kentsanggds (approved):
Tested on integration - LGTM

Issue #1354, opened 2024-11-19
OGL Licence filter doesn't return all results


AgaDufrat:
`license_id` as a facet field counts:

```
"facet_fields":{
      "license_id":[
        "uk-ogl",25775,
        "cc-by",975,
        "OGL-UK-3.0",402,
        "notspecified",119,
        "other-nc",49,
        "cc-zero",17,
        "other-closed",17,
        "other-pd",13,
        "unpublished",11,
        "cc-by-sa",6,
        "other-open",6,
        "odc-odbl",5,
        "odc-by",4,
        "odc-pddl",3,
        "__other__",2,
        "cc-nc",1,
        "ogl",1,
        "other-at",1]}
```
Solr query used - `/solr/ckan/select?fq=license_id%3A"__other__"&indent=true&q.op=OR&q=*%3A*&rows=100`

`uk-ogl` is the code we use to filter datasets that use the standard OGL licence (currently version 3). Datasets with `OGL-UK-3.0` and `ogl` should also be mapped to the same code.

Example query to find individual datasets using a specific licence ID - `/solr/ckan/select?fq=license_id%3A"OGL-UK-3.0"&indent=true&q.op=OR&q=*%3A*&rows=100`

PR #1356, opened 2024-11-20
Ensure OGL Licence filter returns all relevant results

AgaDufrat:
We used `uk-ogl` id to filter datasets that use the standard OGL licence. Datasets with `OGL-UK-3.0` and `ogl` should also be returned.

Available to test in integration https://www.integration.data.gov.uk/search/solr?q= (VPN required)

### Before
<kbd>Screenshot 2024-11-20 at 16 54 07</kbd>

### After
<kbd>Screenshot 2024-11-20 at 16 50 17</kbd>

Closes https://github.com/alphagov/datagovuk_find/issues/1354

### DIFF
--- a/app/presenters/search_presenter.rb
+++ b/app/presenters/search_presenter.rb
@@ -76,6 +76,10 @@ class SearchPresenter
     search_params.dig(:filters, :format)
   end
 
+  def selected_licence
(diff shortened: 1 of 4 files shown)


kentsanggds (approved):
LGTM

PR #1375, opened 2024-12-12
Improve search functionality

AgaDufrat:
### Changes

- Split search terms on whitespaces
- Allow exact phrase search
- Remove stop words from query
- Boost `title` field in query
- Add error handling (incl. `RSolr::Error::Http` error when invalid term is entered)
- _(enhancement)_ Allow advanced search operators (a.k.a special Solr search characters/operators): `+` `-` `&&` `||` `!` `( )` `{ }` `[ ]` `^` `"` `~` `*` `?` `:` `/`. Note: Decided not to escape individual special characters. Searching for special characters. Those also return 'No results found' in OpenSearch.

### Multiple terms
- **Solr before:**  <kbd>after</kbd>

| **Solr after** | **OpenSearch** |
| :---- | :---- |
| after | after |

### Exact phrase search
- **Solr before:**  <kbd>os</kbd>

| **Solr after** | **OpenSearch** |
| :---- | :---- |
| <kbd>after</kbd> | <kbd>os</kbd> |

### Multiple terms incl. exact phrase
- **Solr before:**  <kbd>after</kbd>

| **Solr after** | **OpenSearch** |
| :---- | :---- |
| searches for "animal health" `OR` dogs | appears to search for "animal health" `AND` dogs |
| after</kbd> | <kbd>OS</kbd> |

The same effect can be achieved in Solr when searching for `+"animal health" +dogs` or `"animal health" AND dogs` or `"animal health" && dogs`. Switching Solr to use `AND` operator results in search results being very insistent with OpenSearch. For example `animal health` search results only 220 datasets (instead of over 2000).
 OpenSearch is retuning way less results as it appears to search for "animal health" AND dogs, whereas normally the OR operator is used. After analysing what OpenSearch does I think it’s an odd behaviour as a result of how we process a combination of “phrase” and “term” search, which ends up as and AND query. The standard “term” search (multi_match query) uses the "OR" operator so this seems like an inconsistency in OpenSearch. Users don’t seem to use phase searches anyway so I’d say this is an unlikely scenario and they can also use filters to refine the search. Product decision has been made to accept this difference.

### Testing

Tested with a few popular searches from the [Analytics dashboard](https://lookerstudio.google.com/u/0/reporting/1dLIysu5Ie131ZL3gGT7zyWBhffNgTSX2/page/cduv):
- `organogram of staff roles & salaries` 
  - 1,625 results in Solr (was 41,329 prior to removing stop words)
  - 1,587 results in OpenSearch

- `"organogram of staff roles & salaries”`
  - 273 result in Solr
  - 261 results in OpenSearch

- `ukas certificate`
  - 282 results in Solr
  - 268 results in OpenSearch

- `check numerous`
  - 849 results in Solr
  - 843 results in OpenSearch

### DIFF
--- a/app/controllers/solr_search_controller.rb
+++ b/app/controllers/solr_search_controller.rb
@@ -17,6 +17,8 @@ private
   def solr_search_response
     @solr_search_response ||= begin
       Search::Solr.search(params)
+    rescue Search::Solr::NoSearchTermsError
+      no_results_found
     rescue RSolr::Error::Http => e
       handle_solr_http_error(e)
     end
--- a/app/controllers/solr_search_controller.rb
+++ b/app/controllers/solr_search_controller.rb
@@ -15,6 +15,22 @@ class SolrSearchController < ApplicationController
 private
 
   def solr_search_response
-    @solr_search_response ||= Search::Solr.search(params)
+    @solr_search_response ||= begin
+      Search::Solr.search(params)
+    rescue RSolr::Error::Http => e
+      handle_solr_http_error(e)
+    end
+  end
+
+  def handle_solr_http_error(error)
+    if error.response[:status].to_s.start_with?("4")
+      no_results_found
+    else
+      raise error
+    end
+  end
+
+  def no_results_found
+    { "response" => { "numFound" => 0, "docs" => [] } }
   end
 end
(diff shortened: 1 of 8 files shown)


MuriloDalRi (approved):
Nice work, @AgaDufrat! 👍

PR #1389, opened 2025-01-02
Switch to use Solr

AgaDufrat:
We want to keep existing URLs and since we don't have a feature flag we now have to rename relevant routes, controllers and views.

Renames old routes, controllers and views by adding `os_` prefix and removes `solr_` prefix. Models and Services are kept intact.

Also adds missing translations and _sort partial as well as using Solr for legacy URLs.

- Ensure datafile name falls back correctly to the description

| Before    | After |
| -------- | ------- |
|Screenshot 2025-01-03 at 11 23 55|Screenshot 2025-01-03 at 11 24 00|


kentsanggds (approved):
LGTM

PR #1390, opened 2025-01-03
Use Solr database instead of OpenSearch

AgaDufrat:
To reduce the maintenance of the data.gov.uk stack.


kentsanggds (approved):
Already have done months of testing on Integration and Staging

PR #1396, opened 2025-01-06
Handle missing params

AgaDufrat:
Error occurred because the `params.fetch("q", "")` returns `nil` if q param is missing in the query, and calling squish on nil raises the NoMethodError. params["q"] will return nil if the key doesn't exist, and || will fall back to "" (an empty string) if that's the case.

Also switches to use `.blank?`, which  is more flexible, returning true for objects that are nil, empty, or consist only of whitespace.

https://govuk.sentry.io/issues/6194636580/?project=1461892
```
NoMethodError
undefined method `squish' for nil:NilClass (NoMethodError)

      query_param = params.fetch("q", "").squish
```

### DIFF
--- a/app/presenters/search_presenter.rb
+++ b/app/presenters/search_presenter.rb
@@ -24,7 +24,7 @@ class SearchPresenter
   end
 
   def organisations
-    if search_keywords.empty?
+    if search_keywords.blank?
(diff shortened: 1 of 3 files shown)


kentsanggds (approved):
LGTM

PR #1400, opened 2025-01-08
Increase the limit on organization facet

AgaDufrat:
We were not returning all organisations associated with returned datasets in the Publisher filter because the upper limit on organization facet is too low. It defaults to 100.
https://solr.apache.org/guide/6_6/faceting.html#Faceting-Thefacet.limitParameter

"There are currently around 1500 organisations." was mentioned in
https://github.com/alphagov/datagovuk_find/commit/76b7204cd0c086241ed21589dd728f7c7b4e1aae
We're increasing it to match the [reindexing_script](https://github.com/alphagov/ckanext-datagovuk/blob/4cfb397/ckanext/datagovuk/lib/cli.py#L398)

### DIFF
--- a/app/services/search/solr.rb
+++ b/app/services/search/solr.rb
@@ -1,7 +1,7 @@
 module Search
   class Solr
     RESULTS_PER_PAGE = 20
-    ORGANISATIONS_LIMIT = 1600 # Arbitrary large number
+    ORGANISATIONS_LIMIT = 3000 # Should match `rows` in https://github.com/alphagov/ckanext-datagovuk/blob/4cfb397/ckanext/datagovuk/lib/cli.py#L398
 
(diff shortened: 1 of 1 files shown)


nicholsj:
@AgaDufrat what would be the impact of making the limit even higher - eg 2000 ? That would give more headroom for potential increases and avoid having to touch this again.

AgaDufrat:
> @AgaDufrat what would be the impact of making the limit even higher - eg 2000 (admittedly still arbitrary 😄 )? That would give more headroom for potential increases and avoid having to touch this again.

@nicholsj, Increased to 3000 to match the reindexing script.

nicholsj (approved):
LGTM, thanks!

PR #1405, opened 2025-01-14
Cache organisations for 10 minutes

AgaDufrat:
To improve performance as we're getting a lot of RSolr::Error::ConnectionRefused

Caching is enabled

`config.action_controller.perform_caching = true`

### DIFF
--- a/app/services/search/solr.rb
+++ b/app/services/search/solr.rb
@@ -83,35 +83,39 @@ module Search
     end
 
     def self.get_organisations
-      solr_client = client
-      @organisations_list = {}
+      Rails.cache.fetch("organisations", expires_in: 10.minutes) do
+        solr_client = client
+        @organisations_list = {}
 
-      query = solr_client.get "select", params: {
-        q: "*:*",
-        fq: [
-          "site_id:dgu_organisations",
-        ],
-        fl: %w[title name],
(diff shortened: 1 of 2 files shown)


app/services/search/solr.rb
```diff
@@ -83,35 +83,39 @@ def self.licence_filter(licence)
...
-        q: "*:*",
-        fq: [
-          "site_id:dgu_organisations",
-        ],
-        fl: %w[title name],
-        rows: ORGANISATIONS_LIMIT,
-      }
-      query["response"]["docs"].each do |org|
-        @organisations_list.store(org["title"], org["name"])
+      Rails.cache.fetch("organisations", expires_in: 10.minutes) do
```
theseanything, line 86: Do you know what the current query rate is? You might be able to cache it for even less time and still get decent performance increases. Is there potentially we could cache an error here?

kentsanggds (approved):
LGTM

app/services/search/solr.rb
kentsanggds: Organisations don''t really get updated that often, but at the same time when they are changed it would be nice to see the update soon after the change in Find as changes are now more or less immediate. 

Also I've done some load testing on Integration and there is significant improvement with the 10 minute cache.

PR #1408, opened 2025-01-16
Revert "Only include datasets in the results (exclude harvesters)"

kentsanggds:
Reverts alphagov/datagovuk_find#1407


nicholsj (approved):
LGTM thanks

PR #1426, opened 2025-01-28
Upgrade to rails 7.0

AgaDufrat:
_See commits for details._
~Also includes removal of obsolete code.~


config/application.rb
```diff
@@ -48,8 +48,6 @@ class Application < Rails::Application
 
     config.analytics_tracking_id = ENV["GA_TRACKING_ID"]
     config.i18n.load_path += Dir[Rails.root.join("config/locales/**/**/*.{rb,yml}")]
-    config.filter_parameters << :password
-    config.filter_parameters << :password_confirmation
```
jasonBirchall, line 59: Is `password_confirmation` also passed to the parameter_logging. I can only see `passwd` and that's unclear if it includes both.

jasonBirchall:
This looks good to me. I do have a small question about the filter parameters, but it might be my unfamiliarity with rails.
https://github.com/alphagov/datagovuk_find/pull/1426/commits/739c9ed65b0257ba4bd0b769e5ccd41b9dfeb7e2

A side note. This PR has multiple changes, making it a little hard to track. I suggest separating tech debt changes like the elasticsearch change to its own PR.

config/application.rb
AgaDufrat: My commit message could be a bit more descriptive. Those params were added to configure Raven (Sentry SDK)
https://github.com/alphagov/datagovuk_find/commit/929fda7059ca4df0cc1514c32b0214805763c63e

We migrated over to Sentry Ruby SDK, which protects sensitive data by default so we no longer need to configure those. The default rails configuration is in `config/initializers/filter_parameter_logging.rb`

I'll move this commit to another PR as suggested because it's not directly related to rails 7.0 upgrade.

AgaDufrat:
> This PR has multiple changes, making it a little hard to track. I suggest separating tech debt changes like the elasticsearch change to its own PR.

Yes, sorry. I guess it took an approach to delete unnecessary config instead of adding it back in after running the upgrade script. I removed those commits now and will create a new PR.

jasonBirchall (approved):

PR #1451, opened 2025-03-06
Remove arbitrary app version from build-config.yaml

kentsanggds:
The app version is actually using the release tags on github so remove the app version from the build-config.yaml as they don't add any value.


jasonBirchall:
Moving back to draft until it's ready for review.

PR #1538, opened 2026-01-06
Update privacy.html.erb

antoniaantp:
updating description of what personal data we collect to make clearer it is for publishers using our services. Also updated 'date last updated'

### DIFF
--- a/app/views/pages/privacy.html.erb
+++ b/app/views/pages/privacy.html.erb
@@ -7,7 +7,7 @@
         title: "Privacy"
       } %>
 
-      <p class="govuk-body">Last updated 21 July 2025</p>
+      <p class="govuk-body">Last updated 6 Jan 2026</p>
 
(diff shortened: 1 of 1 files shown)


app/views/pages/privacy.html.erb
```diff
@@ -61,7 +61,7 @@
       } %>
 
       <p class="govuk-body">
-        We collect your personal data so we can provide access to the publishing
+        We collect personal data from data publishers using data.gov.uk to publish their data so we can provide access to the publishing
```
Obsiye: hi, i had a look at the privacy page and it shows:
```
What data we need
The personal data we collect from you will include:
...
```
does this also need updating ? because it says the personal data we collect from **you**

antoniaantp:
@kentsanggds can you push this for me?

Obsiye (approved):

PR #1541, opened 2026-01-08
DGUK-115 Validate search params

Obsiye:
Validate search params

1. Prevent non English/Welsh characters at the moment whilst we resolve bot attempts on the search query.
2. Test non english and welsh letters get redirected to the home page

**Decision made:**
I did not provide an error message just in case a bot will retry. Happy to implement the error summary, if needed.

3. Add Gem rack-attack
4. Add request rate throttling to block a bot's ip address. (block ip's when 300 requests made in 60 seconds)
5. Test rack-attack throttling

### DIFF
--- a/app/controllers/search_controller.rb
+++ b/app/controllers/search_controller.rb
@@ -1,18 +1,32 @@
 class SearchController < ApplicationController
+  before_action :validate_search_params!
+
+  LATIN_AND_NUMBERS_REGEX = /[\p{Latin}\p{N}\s]/
+  SPECIAL_CHARS_REGEX = /[£$!%&'*+_=\-.\/,:;"`?<>@#^(){}\[\]~|\\]/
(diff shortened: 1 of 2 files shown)


config/initializers/rack_attack.rb
```diff
@@ -0,0 +1,10 @@
+class Rack::Attack
+  RATE_LIMIT_COUNT = ENV["RATE_LIMIT_COUNT"]&.to_i || 300
+  RATE_LIMIT_PERIOD = (ENV["RATE_LIMIT_PERIOD_SECONDS"]&.to_i || 60).seconds
+
+  throttle("/search", limit: RATE_LIMIT_COUNT, period: RATE_LIMIT_PERIOD) do |request|
```
kentsanggds: I wonder if there is a way of introducing a test just so that we know what path is being throttled? This would make it clear in the tests that we are only targeting `/search`

Obsiye: hi i can give that a try now

kentsanggds: Thanks, I find tests a good way of getting an understanding of what the code is doing without actually reading through the code

PR #1542, opened 2026-01-14
DGUK-115 Reduce throttling threshold

Obsiye:
Reduce the default rack-attack threshold to 120 requests per minute per ip address / user

This will help us block bots IP addresses from overwhelming our service / search feature.

The decision to set to 120 requests per minute per IP is an initial assumption. A user wouldn't normally request that many per minute. However, this is a trial and we'll adjust this value.


kentsanggds (approved):
Looks fine, we'll start with a rate limit of 120 for a single IP., and can see how that goes

Can you tidy up the commits and mention reason for the limit

PR #1557, opened 2026-02-20
add frontend docs

joelanman:
Add a guide to frontend development for v2, obviously can be updated as we go

### DIFF
--- a/docs/frontend.md
+++ b/docs/frontend.md
@@ -1,0 +1,41 @@
+# Frontend development guide for v2
+
+Templates folder
+
+```
+app/views/v2
+```
+
+Assets folders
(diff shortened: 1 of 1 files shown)


docs/frontend.md
```diff
@@ -0,0 +1,41 @@
...
+
+Assets folders
+
+```
+app/assets/images/v2
+app/assets/javascripts/v2
+app/assets/stylesheets/v2
+```
+
+## Stylesheets
```
currycoder: Had a quick thought just now.  Is it worth calling out the `/components` test page here?  That when adding new styling/components we should make sure they appear here?

PR #1574, opened 2026-02-27
[DGUK-105] Data manual home

currycoder:
This change adds a data manual home page as follows;

image

### DIFF
--- a/spec/features/collections_spec.rb
+++ b/spec/features/collections_spec.rb
@@ -22,7 +22,7 @@ RSpec.feature "collections", type: :feature do
 
   def and_i_can_see_the_feedback_form
     expect(page).to have_css(".govuk-inset-text.datagovuk-feedback-inset-text")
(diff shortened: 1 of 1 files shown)


app/views/v2/components/_cta.html.erb
joelanman: do we still need this?

app/views/v2/components/_feedback.html.erb
```diff
@@ -0,0 +1,10 @@
+<div class="govuk-inset-text datagovuk-feedback-inset-text">
+  <p class="govuk-body datagovuk-body">
+    <% if local_assigns.fetch(:dataset, false) %>
```
Obsiye: could this be ` if local_assigns[:dataset]` ? it just reads better for me. 
```
<% if local_assigns[:dataset] %>
  Is this data/dataset useful?
<% else %>
  Is this page useful?
<% end %>
```

PR #1577, opened 2026-03-02
[DGUK-210] Data manual content

currycoder:
This change adds the data manual content pages as follows;
image

The content for these pages is sourced/rendered from markdown files in `app/contents/data_manual/` matching the URL's `slug` parameter.  This means that the markdown is rendered per-request to the rails application.  This breaks from the precedent set for the collection pages where we rendered the markdown as part of a rake task and saved static HTML. 
Some local profiling tells us the "render on request" approach adds 10-20ms to the time to service the request which feels acceptable given the much improved developer experience.

~**TODO:** Point this PR at main-mvp when the parent branch merges.~

### DIFF
--- a/app/controllers/v2/data_manual_controller.rb
+++ b/app/controllers/v2/data_manual_controller.rb
@@ -5,7 +5,6 @@ module V2
   end
 
   class DataManualController < ApplicationController
-
     rescue_from DataManualContentNotFound, with: :render_not_found
 
     def home
@@ -22,7 +21,7 @@ module V2
       }
     end
 
-    private
+  private
 
     def render_content
       # The sanitizer is a double protection against attempts to render markdown files
@@ -40,8 +39,7 @@ module V2
       else
         raise DataManualContentNotFound
       end
-      rendered_content = Dgu::Markdown.render(markdown).html_safe
-      return rendered_content
+      Dgu::Markdown.render(markdown).html_safe
     end
 
     def data_manual_pages
@@ -51,7 +49,7 @@ module V2
           data_manual_item[:selected] = true
         end
       end
-      return data_manual_pages
+      data_manual_pages
     end
   end
 end


app/assets/stylesheets/v2/components/_section_navigation.scss
```diff
@@ -0,0 +1,13 @@
+.datagovuk-section-navigation .datagovuk-section-navigation__item {
+  padding: 10px;
```
joelanman: can we change the approach so this is

padding: 10px 0;

then the selected is:

margin-left: -15px;
padding-left: 11px;

pulling the selected line out - that way the grid lines up for the text

currycoder: I found keeping padding as-is on this one and changing selected to...

```
  margin-left: -10px;
  padding-left: 11px;
```

...rendered this fairly nicely on mobile and desktop and kept the alignment with "Get in touch".

joelanman: actually, selected doesnt look aligned in your screenshot?

app/controllers/v2/data_manual_controller.rb
```diff
@@ -1,8 +1,42 @@
...
+      expires_in 30.minutes, public: true
+      # The sanitizer is a double protection against attempts to render markdown files
+      #   other than the ones in the data manual. The route also specifies the slug parameter
+      #   as having the slug constraint
+      @sanitizer = Rails::Html::FullSanitizer.new
+      safe_slug = @sanitizer.sanitize(params[:slug])
+      data_manual_markdown_directory = Rails.configuration.x.markdown_data_manual_location
+      # NOTE: the below brakeman complaint is happening because brakeman cannot verify that our config
+      #   variable does not contain a leading /.  Ignoring it.
+      # brakeman: disable: Rails/DynamicRenderPath, Rails/FileAccess
```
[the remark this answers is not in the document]
currycoder: This is done.  I had to do a little bit of exception handling to still achieve the same 404 response given the use of helper methods

PR #1584, opened 2026-03-04
DGUK-241 create headline chart

Obsiye:
Add a headline chart
Add images (up and down arrow)
Add stylesheets
Place headline chart in the components page

image


joelanman (changes requested):
the layout isnt quite right - it's breaking out of the column, I'll try another approach to the layout and update

currycoder (approved):

PR #1596, opened 2026-03-10
[DGUK-221] Add correct google analytics domains to CSP settings

currycoder:
Currently on data.gov.uk production, we can see errors in the browser console as follows;
```
Connecting to 'https://region1.google-analytics.com/g/collect?v=2' violates the following Content Security Policy directive: "connect-src 'self' www.google-analytics.com". The action has been blocked.

Fetch API cannot load https://region1.google-analytics.com/g/collect?v=2.... Refused to connect because it violates the document's Content Security Policy.
```

It seems that we are only receiving partial google analytics data - this tallies with what we see in the analytics dashboards as certain geographical regions are suspiciously low on pageviews.

This change fixes the issue by allowing the domains which are required by google analytics in the app's content security policy.

### DIFF
--- a/config/initializers/secure_headers.rb
+++ b/config/initializers/secure_headers.rb
@@ -2,13 +2,13 @@ SecureHeaders::Configuration.default do |config|
   config.csp = {
     preserve_schemes: true,
     default_src: ["'none'"],
-    connect_src: ["'self'", "www.google-analytics.com"],
+    connect_src: ["'self'", "*.google-analytics.com", "*.googletagmanager.com", "*.analytics.google.com"],
     font_src: ["'self'", "data:"],
-    img_src: ["'self'", "www.google-analytics.com"],
+    img_src: ["'self'", "*.google-analytics.com", "*.googletagmanager.com"],
     manifest_src: ["'self'"],
     media_src: ["'self'"],
     object_src: ["'self'"],
-    script_src: ["'unsafe-inline'", "'self'", "www.google-analytics.com", "www.googletagmanager.com"],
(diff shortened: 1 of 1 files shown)


ashimali (approved):

Nooshu:
Although this fix works, the use of `unsafe-inline` isn't ideal. I'd recommend using a [nonce](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Global_attributes/nonce) instead. The ideal CSP would look like this:

```
Content-Security-Policy:
  default-src 'self';
  script-src 'self' 'nonce-<RANDOM>' https://www.googletagmanager.com;
  connect-src 'self' https://www.google-analytics.com https://region1.google-analytics.com https://region2.google-analytics.com;
  img-src 'self' data: https://www.google-analytics.com;
  style-src 'self';
  frame-src https://www.googletagmanager.com;
```
The nonce should be:

- Random (not guessable)
- Unique per request
- Base64 encoded
- Typically at least 128 bits (16 bytes)

I hope that helps! Keep up the great work! I love that data is back and being actively developed and improved! 😍

currycoder:
Thanks for the explainer @Nooshu!  Agreed about `unsafe-inline`.  I wanted to keep this PR pretty light touch, so it fixes the google analytics issue we were experiencing and leaves things at that.

We are now working on a python stack for this app over at https://github.com/alphagov/datagovuk

We're in the process of adding `nonce` support here; https://github.com/alphagov/datagovuk/pull/2/changes/0274677b7f48018d37e09dbdb85e528c82f8bcac

PR #1600, opened 2026-03-10
[DGUK-272] Style V2 cookie banner

currycoder:
This adds V2 styling to the cookie banner;
image

I've broken the HTML out from the govuk publishing components library so that we have more control over styling.  Much like the cookies page https://github.com/alphagov/datagovuk_find/pull/1599 it's a bit of a half way house towards our ideal standards because we are limited by the javascript which we don't really want to touch at this stage.


ashimali (approved):
Looks good to me. Reason for departing from conventions captured in comment, so 👍🏼

PR #1618, opened 2026-03-13
DGUK-262 Add govuk govspeak barchart

Obsiye:
Add govuk [magnacharta](https://github.com/alphagov/magna-charta) bar chart

This uses code found in the Gemfile govuk_publishing_components

image

Optional: with the "Change to table" button
image


Obsiye:
we have highcharts license now

PR #1633, opened 2026-03-17
[DGUK-270] Collection pagination

currycoder:
This change adds simple pagination to collections pages as follows;
image

image

### DIFF
--- a/app/services/dgu/collections_service.rb
+++ b/app/services/dgu/collections_service.rb
@@ -60,9 +60,9 @@ module Dgu
     end
 
     def previous_page
-     return if current_page_index <= 0
-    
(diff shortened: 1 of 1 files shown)


app/assets/stylesheets/v2/components/_collection_pagination.scss
```diff
@@ -0,0 +1,47 @@
+.datagovuk-collection-pagination__card {
+  background-color: $datagovuk-light-green;
+  border-radius: 30px;
+  padding-right: 0;
+  margin-bottom: 2em;
```
joelanman: for the most part, we should do text in relative units (rem) and most other measurements in px. That way large font size doesnt make all the gaps and stuff massive

PR #1843, opened 2026-03-30
DGUK-290 Remove generated views by markdown 

Obsiye:
- Removed app/views/generated directory
- Docker now builds the generated views as an output of the markdown render rake task
- Runs markdown render **before the whole feature tests** and then deletes it


ashimali (approved):
I would just add a bit of clarification to readme about requirement of local rendering.

I know we may move soon and we might all know this but in case another dev wanted to checkout and see things working as they should, then I think useful for the readme to be clear on that point.

ashimali (approved):

ashimali (approved):

PR #1864, opened 2026-04-20
Update the site_id field which is used to store an index of organisation

kentsanggds:
The index of organisations on production grew beyond the 3000 row limit, there are less than 1500 orgs on the system so the 3000 limit should give enough headroom to handle extra organisations for a few years. Rather than purge the existing index and repopulate it which will cause some downtime with the directory, a new index has been created which has the correct number of publishers. 

So point the code to the new organisation index.

https://cddodatamarketplace.atlassian.net/jira/software/c/projects/DGUK/boards/527?selectedIssue=DGUK-432

### DIFF
--- a/app/services/search/solr.rb
+++ b/app/services/search/solr.rb
@@ -2,6 +2,7 @@ module Search
   class Solr
     RESULTS_PER_PAGE = 20
     ORGANISATIONS_LIMIT = 3000 # Should match `rows` in https://github.com/alphagov/ckanext-datagovuk/blob/4cfb397/ckanext/datagovuk/lib/cli.py#L398
+    DGU_ORGANISATIONS_ID = "dgu_organisations_2".freeze # Should match DGU_ORGANISATIONS_ID in https://github.com/alphagov/ckanext-datagovuk/blob/e774ed637178533a47ddfa8fadaf83adc9e903be/ckanext/datagovuk/lib/cli.py#L26
 
     def self.search(params)
       query_param = (params["q"] || "").to_s.squish
@@ -24,7 +25,7 @@ module Search
       processed_query = SearchHelper.process_query(query_param)
       raise NoSearchTermsError, "Query string is empty after processing" if processed_query.blank?
 
-      @query = "(title:(#{processed_query})^2 OR notes:(#{processed_query})) AND NOT site_id:dgu_organisations"
+      @query = "(title:(#{processed_query})^2 OR notes:(#{processed_query})) AND NOT site_id:#{DGU_ORGANISATIONS_ID}"
     end
 
     def self.build_filter_query(params)
@@ -92,7 +93,7 @@ module Search
         query = solr_client.get "select", params: {
           q: "*:*",
           fq: [
-            "site_id:dgu_organisations",
+            "site_id:#{DGU_ORGANISATIONS_ID}",
           ],
           fl: %w[title name],
           rows: ORGANISATIONS_LIMIT,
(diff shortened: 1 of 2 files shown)


antoniaantp (approved):
Looks good

currycoder:
How did the index grow to this size if there's only ~1500 orgs?

app/services/search/solr.rb
```diff
@@ -2,6 +2,7 @@ module Search
   class Solr
     RESULTS_PER_PAGE = 20
     ORGANISATIONS_LIMIT = 3000 # Should match `rows` in https://github.com/alphagov/ckanext-datagovuk/blob/4cfb397/ckanext/datagovuk/lib/cli.py#L398
+    DGU_ORGANISATIONS_ID = "dgu_organisations_2" # Should match DGU_ORGANISATIONS_ID in https://github.com/alphagov/ckanext-datagovuk/blob/e774ed637178533a47ddfa8fadaf83adc9e903be/ckanext/datagovuk/lib/cli.py#L26
```
currycoder: Can you explain why we needed to change this?  As I understand it we only need to change the SOLR index itself which I would assume would be captured in the `SOLR_URL` environment variable

currycoder (commented):
Some questions inline

app/services/search/solr.rb
```diff
@@ -24,7 +25,7 @@ def self.build_term_query(query_param)
       processed_query = SearchHelper.process_query(query_param)
       raise NoSearchTermsError, "Query string is empty after processing" if processed_query.blank?
 
-      @query = "(title:(#{processed_query})^2 OR notes:(#{processed_query})) AND NOT site_id:dgu_organisations"
+      @query = "(title:(#{processed_query})^2 OR notes:(#{processed_query})) AND NOT site_id:#{DGU_ORGANISATIONS_ID}"
```
currycoder: I don't really understand the `AND NOT site_id:dgu_organisations_2` bit in general...  What are we achieving with it?

app/services/search/solr.rb
kentsanggds: I'm not reindexing the whole CKAN index, this is an index just for organisations which is used by Find to quickly get organisation details rather than performing lookups based on an org id

app/services/search/solr.rb
kentsanggds: The organisations index is actually reusing a dataset schema, I made a decision a while ago to reuse this rather than trying to create a new schema just for organisation because this would mean that the datagovuk wouldn't diverge too much from the core CKAN solr schema. We might want to revisit this decision in the future but for now it works without interfering with other directory searches. 
I might have to change this slightly to check that it doesn't start with `dgu_organisations` rather than being specific

currycoder: Ah ok understood.  So documents with `site_id:#{DGU_ORGANISATIONS_ID}` are explicitly *not* for searching by datagovuk find - that is a marker that separates documents from something CKAN uses to something datagovuk is allowed to search upon.  Is my understanding correct?

Yes I feel like excluding things where `site_id` begins with `dgu_organisations` could be more bulletproof.  We would avoid unexpected results while both "sub-indexes" are present in this index.

kentsanggds:
> How did the index grow to this size if there's only ~1500 orgs?

I've got no idea, on Integration it is still only about 1477, on Staging it is almost 3000, there is probably some issue with the solr results not returning fully sometimes, causing an outage leading to additional publishes to be added in. There is probably another ticket to make it more robust by checking that the org doesn't exist just before it gets added in.

PR #1889, opened 2026-05-20
DGUK-317: Header styling issue fixed

sho88:
# Current PR

**What:** 
- Header adjusted according to the screen widths
- Spacing in the menu heading has been corrected

**Why:** 
- To be consistent with the rest of the data.gov.uk website...

### DIFF
--- a/app/assets/stylesheets/v2/components/_header.scss
+++ b/app/assets/stylesheets/v2/components/_header.scss
@@ -7,6 +7,12 @@
     flex-wrap: wrap;
     align-items: start;
     gap: 0 20px;
+    margin: auto;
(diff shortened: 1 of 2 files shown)


currycoder (commented):
I think these fixes targeted the header when viewed through `/directory`, but I think they make things go a little bit off when applied to the V2 pages (homepage, collections, etc)

The problem is that we are reusing the same header (components/styling etc) on both V2 and directory.

Would recommend holding off merging this in until we are fully moved over to the python app serving the V2 endpoints.  That way there would be no collision.

joelanman (changes requested):
can we try the media query approach

PR #1907, opened 2026-06-05
[DGUK-497] Drop V2 related content/controllers/views/tests

currycoder:
We have now superseded the V2/MVP pages contained within this project in the https://github.com/alphagov/datagovuk python application.

This change drops most of the V2/MVP related code for the find app;
- Content markdowns.  This is the primary aim of the PR, to ensure we have markdown defined only in a single place.  Companion PR which adds the markdown to `datagovuk` repo; https://github.com/alphagov/datagovuk/pull/143
- V2 controllers/views for; home page, collections pages, data manual pages, peripheral pages.
- Services; markdown/collections rendering services.
- Tasks; markdown render task.

It **does not** drop any of the V2 styling.  Nor does it drop the V2 header component, error pages or components page.  The header/styling is a little more tricky to unpick as we are using it on the directory pages - so this will be saved for a follow-up PR.


docker/dev.Dockerfile
```diff
@@ -15,12 +13,9 @@ RUN apt-get update && apt-get install -y \
...
 RUN bundle exec rails assets:precompile
-RUN bundle exec rake markdown:render
 
 # Ubuntu 24.04 will always use snap to install chromium 
 # so use an alternative PPA to install chromiumn and chromiun-driver
 RUN add-apt-repository ppa:xtradeb/apps -y && \
     apt update && \
     apt install -y chromium chromium-driver
-
-USER app
```
currycoder, line 26: Removed this as it trips up the docker-compose stack when re-starting the rails server after file changes

ashimali (approved):

app/views/datasets/_no_datafiles.html.erb
```diff
@@ -5,8 +5,8 @@
       <%= t('datasets.show.contact_the_publisher') %>
     <% else %>
       <%= t('datasets.show.contact_the_team') %>
-      <%= link_to 'data.gov.uk/support', support_path, class: 'govuk-link' %>
+      <%= link_to 'data.gov.uk/support', "/support", class: 'govuk-link' %>
```
Obsiye: feel free to ignore this since this app is eventually sunsetting.

You could use a direct route in `config/routes.rb`

https://guides.rubyonrails.org/routing.html#direct-routes

e.g.

```
  direct :support do
    "/support"
  end
```

and return to using `support_path` in the views

config/routes.rb
```diff
@@ -1,6 +1,6 @@
 Rails.application.routes.draw do
   # For details on the DSL available within this file, see http://guides.rubyonrails.org/routing.html
-  root to: "v2/pages#home"
+  root to: redirect(path: "/search")
```
Obsiye, line 3: about this. when in the `app/views/v2/_header.html.erb` and clicking the logo `national data library` it redirects to the directory page. i'm assuming it should redirect to the python app instead?


maybe the only thing that needs updating is the `_header.html.erb` to be a direct link to `/` ? assuming that then redirects to the python app home page




Image

currycoder: Yeah this is just an oddity of our traffic split I think.  The href for the logo is to "/" as you say, but I've just made it so that `/` on rails find redirects to `/search`

In practice on integration/staging/production this would link to the homepage of the python app.

We could maybe look at making our local docker stack a bit more intelligent and match the production traffic split more closely.

app/views/datasets/_no_datafiles.html.erb
currycoder: Thanks for flagging this - will take a look

config/routes.rb
Obsiye: ohhh got it makes sense thanks

Obsiye (approved):

